PR #5, opened 2023-08-30
DCMAW-6087 GitHub workflows

cgedrim-co:
### Add workflows and actions

- Workflows
  - on_push
  - on_pull_request
- Actions
  - bundle-reports
  - detect-os
  - gradle-check
  - gradle-connected-test
  - setup-runner

Resolves: DCMAW-6087


.github/workflows/on_pull_request.yml
```diff
@@ -0,0 +1,52 @@
+name: On Pull Request (develop)
+
+on:
+  pull_request:
+    branches:
+      - develop
```
Chuddington, line 6: As part of following the gitflow branching strategy, we may want to include main in here as well

config/actions/detect-os/action.yml
Chuddington, line 1: Do we want to provide outputs for this action, as opposed to appending the values to the environment?

https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#creating-an-action-metadata-file

.github/workflows/on_pull_request.yml
cgedrim-co: Yea, I was going to clarify that with Oli before I put it in

config/actions/setup-runner/action.yml
Chuddington, line 1: I'll just put it here as it seems to be quite common - we may want to split out linux / mac environment setup. As well as using provided inputs. Although, we wouldn't necessarily need inputs / environment variables if we have two separate actions that are then conditional based on the action's output

app/src/androidTest/java/uk/gov/onelogin/ExampleInstrumentedTest.kt
```diff
@@ -6,19 +6,14 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
...
- * See [testing documentation](http://d.android.com/tools/testing).
- */
 @RunWith(AndroidJUnit4::class)
 class ExampleInstrumentedTest {
     @Test
     fun useAppContext() {
         // Context of the app under test.
         val appContext = InstrumentationRegistry.getInstrumentation().targetContext
-        assertEquals("com.example.onelogin", appContext.packageName)
+        assertEquals("uk.gov.onelogin.staging", appContext.packageName)
```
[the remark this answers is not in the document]
cgedrim-co: Yup, but I've set the workflow to only run staging tests (as per instruction), and I didn't want to muddy the PR waters by adding flavor-specific tests in for now

config/actions/git-tag-bundles/action.yml
```diff
@@ -0,0 +1,46 @@
+name: 'Get the latest annotated tag'
+description: 'Gets the latest tag on a branch'
+
+inputs:
+  abb-paths:
+    description: 'Paths to build flavor abb'
```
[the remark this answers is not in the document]
cgedrim-co: That's just my lysdexia... it should be `aab` 🤦

config/actions/git-tag-bundles/action.yml
```diff
@@ -0,0 +1,46 @@
...
+  using: "composite"
+  steps:
+    - name: Git tag bundles
+      run: |
+        BUNDLES=($(echo $INPUT_ABB_PATHS | tr ";" "\n"))
+        
+        for BUNDLE in ${FLAVORS[@]}
+        do
+          IFS=- read -r FLAVOR ABB_PATH <<< $BUNDLE
+
```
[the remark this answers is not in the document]
cgedrim-co: They're just strings that I manually pull apart in bash, so yes

Chuddington (approved):

config/styles/Vocab/Base/reject.txt
chrisdondig, line 1: Is this empty file needed?

chrisdondig (approved):

config/styles/Vocab/Base/reject.txt
cgedrim-co: Yes? I think that's one that @Chuddington added, but it's vocab for Vale

PR #25, opened 2023-09-29
Feature/dcmaw 6343 create pre login splash screen

de1phine:
Create pre login splash screen


app/src/main/java/uk/gov/onelogin/MainActivity.kt
```diff
@@ -1,58 +1,41 @@
...
-    override fun onOptionsItemSelected(item: MenuItem): Boolean {
-        // Handle app bar item clicks here. The app bar
-        // automatically handles clicks on the Home/Up button, so long
-        // as you specify a parent activity in AndroidManifest.xml.
-        return when (item.itemId) {
-            R.id.action_settings -> true
-            else -> super.onOptionsItemSelected(item)
-        }
-    }
+class MyViewModel : ViewModel() {
```
cgedrim-co, line 30: All this model seems to be doing is delaying for 2 seconds, is this just sample code?

If we're never intending to release with this in here I'd suggest making the code as pared down as possible to save "test" code from being inadvertently kept.

app/src/main/java/uk/gov/onelogin/Theme.kt
```diff
@@ -0,0 +1,54 @@
...
+    onSurface = md_theme_light_onSurface,
+    primary = md_theme_light_primary,
+    primaryVariant = md_theme_light_primaryVariant,
+    secondary = md_theme_light_secondary,
+    secondaryVariant = md_theme_light_secondary,
+    surface = md_theme_light_surface
+)
+
+@Composable
+fun GdsTheme(
```
cgedrim-co, line 40: For anyone else reading this PR: `GDS` in this instance is Government Design System

app/src/main/res/values-night/colors.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="icon_background">#1D70B8</color>
```
cgedrim-co, line 3: As this value doesn't change between default and night I think you can safely omit it

app/src/main/res/values/splash.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="Theme.App.First" parent="Theme.SplashScreen">
+        <item name="windowSplashScreenBackground">#1D70B8</item>
```
cgedrim-co, line 5: Should this reference `icon_background`?

app/src/main/res/values/themes.xml
```diff
@@ -6,4 +6,22 @@
...
+        <!-- Set the splash screen background, animated icon, and animation
+        duration. -->
+        <item name="windowSplashScreenBackground">@color/icon_background</item>
+
+        <!-- Use windowSplashScreenAnimatedIcon to add a drawable or an animated
+             drawable. One of these is required. -->
+        <item name="windowSplashScreenAnimatedIcon">@drawable/ic_launcher_foreground</item>
+
+        <!-- Required for animated icons. -->
+<!--        <item name="windowSplashScreenAnimationDuration">200</item>-->
```
cgedrim-co, line 20: Adding this line back in, and removing the viewModel delay might be the best course of action for now

app/src/main/java/uk/gov/onelogin/Theme.kt
Chuddington: Another thing to mention is that this file would be removed once the UI framework is included as a dependency within the `build.gradle.kts` files

cgedrim-co (approved):

PR #32, opened 2023-11-06
DCMAW-6344

cgedrim-co:
Use `LandingScreen` page from ui common repo
Hook up primary call to action to open webview

https://github.com/alphagov/di-mobile-android-onelogin-app/assets/113696444/ea61747d-1a78-4fec-a72d-5362da06092d


MohamedHassan101 (approved):

.github/workflows/on_pull_request.yml
```diff
@@ -57,21 +57,27 @@ jobs:
...
           version-code: ${{ steps.version-code.outputs.version-code }}
           version-name: ${{ steps.version-name.outputs.version-name }}
 
       - name: Run gradle instrumentation tests
         uses: ./config/actions/gradle-connected-test
         with:
+          access-token: ${{ secrets.PACKAGES_PAT }}
           architecture: 'arm64-v8a'
           device-profile: 'pixel_xl'
+          username: ${{ secrets.PACKAGES_USER }}
```
elishkaflint, line 71: Could move this up next to access-token so it's consistent with other steps?

.github/workflows/on_push-release.yml
```diff
@@ -97,8 +97,10 @@ jobs:
         id: bundle-release
         uses: ./config/actions/gradle-assemble-and-bundle
         with:
+          access-token: ${{ secrets.PACKAGES_PAT }}
           flavors: 'staging,production'
 #          flavors: 'staging,production'
```
elishkaflint, line 102: could boy scout and remove this comment while we're here?

.github/workflows/on_pull_request.yml
elishkaflint: Applies to other steps in this file as well

.gitignore
```diff
@@ -358,3 +358,4 @@ obj/
 !/gradle/wrapper/gradle-wrapper.jar
 
 # End of https://www.toptal.com/developers/gitignore/api/java,gradle,kotlin,android,intellij+all,androidstudio,macos,windows
+github.properties
```
elishkaflint, line 361: @cgedrim-co what is in your `github.properties` file and how comes it's ignored?

app/src/main/java/uk/gov/onelogin/MainActivity.kt
```diff
@@ -5,6 +5,7 @@ import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.navigation.compose.rememberNavController
+import uk.gov.ui.theme.GdsTheme
```
elishkaflint, line 8: How come we need this import? How is it used here?

build.gradle.kts
```diff
@@ -1,3 +1,7 @@
+
```
elishkaflint, line 1: Learning question - how is this build.gradle.kts different from the one that's in the `app` directory?

local.properties.example
```diff
@@ -0,0 +1,8 @@
+## This file must *NOT* be checked into Version Control Systems,
```
elishkaflint, line 1: Should this be checked in given this comment...?

.gitignore
cgedrim-co: It's used locally to pull packages from the github maven repo

app/src/main/java/uk/gov/onelogin/MainActivity.kt
cgedrim-co: GdsTheme is used to skin the Material 3 components for the whole app

local.properties.example
cgedrim-co: Yes, this file should be checked in as it's the `.example` file, it should be renamed without the suffix and is used to set a local versionCode and versionName

.github/workflows/on_pull_request.yml
cgedrim-co: this is already consistent as it's alphabetised because alphabetising is next to godliness

.github/workflows/on_push-release.yml
cgedrim-co: Not entirely sure why that comment is there.... I'll remove it

build.gradle.kts
Chuddington: Yeah so the different `build.gradle.kts` in each directory shows the module structure within the root project `build.gradle.kts`.

This in turn means that within a given Android gradle project, you will often have an `:app` submodule (with it's own `build.gradle.kts`), along with any other gradle submodules made, such as `:modules:readid:common`.

.github/workflows/on_push-release.yml
cgedrim-co: Actually... that comment _should_ remain; the line above needs to change to:

```
flavors: 'staging'
```

.github/workflows/on_pull_request.yml
elishkaflint: you are right!

local.properties.example
elishkaflint: That comment is a bit misleading then...

.github/workflows/on_push-release.yml
cgedrim-co: ...because we've not _yet_ got the production app setup in the play store console (it really helps if I remember to add context!)

Chuddington (approved):

PR #33, opened 2023-11-07
Chore: update to UI v2.0.0

cgedrim-co:
Resolves: DCMAW-6344

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -41,7 +41,6 @@ android {
             signingConfig = signingConfigs.getByName("release")
         }
         debug {
-            compileSdkPreview = "UpsideDownCake"
             enableUnitTestCoverage = true
             enableAndroidTestCoverage = true
         }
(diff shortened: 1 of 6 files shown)


app/build.gradle.kts
```diff
@@ -43,6 +41,7 @@ android {
             signingConfig = signingConfigs.getByName("release")
         }
         debug {
+            compileSdkPreview = "UpsideDownCake"
```
Chuddington, line 44: If it's only going to be for debug builds, I'm hesitant that we are inadvertently building something that would be used before the general public.

What I mean by this is that we would only be verifying the version with this flag set, so if there's an issue with the non-debug version, I'm not sure there are tests that help verify that.

With this in mind, is there an issue with removing this entirely then updating the various `compileSdk`, `targetSdk` and the `minimumSdk` properties?

MohamedHassan101 (approved):
Looks good

PR #35, opened 2023-11-08
DCMAW-7076:Android | Initiate login

MohamedHassan101:
### DCMAW-7076:Android | Initiate login
- Update url build with the correct query parameters
- Update url staging with the correct query parameters

Video evidence for **Staging** 

https://github.com/govuk-one-login/mobile-android-one-login-app/assets/113670948/a68d4003-db43-4181-b5a0-9b33fefb3f52

### DIFF
--- a/app/src/build/res/values/login_flow.xml
+++ b/app/src/build/res/values/login_flow.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="openIdConnectBaseUrl" translatable="false">https://auth-stub.mobile.dev.account.gov.uk%1$s</string>
-    <string name="openIdConnectClientId" translatable="false">CLIENT_ID</string>
+    <string name="openIdConnectClientId" translatable="false">TEST_CLIENT_ID</string>
     <string name="webBaseUrl" translatable="false">https://mobile-build.account.gov.uk%1$s</string>
 
 
--- a/app/src/build/res/values/login_flow.xml
+++ b/app/src/build/res/values/login_flow.xml
@@ -3,7 +3,4 @@
     <string name="openIdConnectBaseUrl" translatable="false">https://auth-stub.mobile.dev.account.gov.uk%1$s</string>
     <string name="openIdConnectClientId" translatable="false">TEST_CLIENT_ID</string>
     <string name="webBaseUrl" translatable="false">https://mobile-build.account.gov.uk%1$s</string>
-
-
-
 </resources>
--- a/app/src/build/res/values/login_flow.xml
+++ b/app/src/build/res/values/login_flow.xml
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="openIdConnectBaseUrl" translatable="false">https://oidc.build.account.gov.uk%1$s</string>
+    <string name="openIdConnectBaseUrl" translatable="false">https://auth-stub.mobile.dev.account.gov.uk%1$s</string>
     <string name="openIdConnectClientId" translatable="false">CLIENT_ID</string>
     <string name="webBaseUrl" translatable="false">https://mobile-build.account.gov.uk%1$s</string>
+
+
+
 </resources>
--- a/app/src/main/java/uk/gov/onelogin/login/UriBuilder.kt
+++ b/app/src/main/java/uk/gov/onelogin/login/UriBuilder.kt
@@ -12,7 +12,7 @@ data class UriBuilder(
 ) {
   val url: Uri = Uri.parse(baseUri)
             .buildUpon().appendQueryParameter("response_type", "code")
-            .appendQueryParameter("scope", "openid email phone offline_access")
+            .appendQueryParameter("scope", "openid offline_access")
             .appendQueryParameter("client_id", clientID)
             .appendQueryParameter("state", state)
             .appendQueryParameter(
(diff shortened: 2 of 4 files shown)


app/src/staging/res/values/login_flow.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="openIdConnectBaseUrl" translatable="false">https://oidc.staging.account.gov.uk%1$s</string>
-    <string name="openIdConnectClientId" translatable="false">CLIENT_ID</string>
-    <string name="webBaseUrl" translatable="false">https://mobile-staging.account.gov.uk%1$s</string>
-</resources>
\ No newline at end of file
+    <string name="openIdConnectBaseUrl" translatable="false">https://oidc.integration.account.gov.uk%1$s</string>
+    <string name="openIdConnectClientId" translatable="false">sdJChz1oGajIz0O0tdPdh0CA2zW</string>
```
Chuddington, line 4: I don't know if it matters so I'll ask anyway - are we happy with having this hardcoded within the strings / public code base?

Chuddington (approved):
Looks good. Got a comment on the Client ID

app/src/staging/res/values/login_flow.xml
cgedrim-co: @elishkaflint one for you maybe?

elishkaflint: Thanks for flagging @Chuddington! I'm checking and will come back on this as soon as possible. I think the client id is a public identifier and not considered a secret per the spec (there are other controls eg. the client assertion - signed with a secret private key - during the token exchange) but let me double-check that.

app/src/main/java/uk/gov/onelogin/login/UriBuilder.kt
```diff
@@ -12,7 +12,7 @@ data class UriBuilder(
 ) {
   val url: Uri = Uri.parse(baseUri)
             .buildUpon().appendQueryParameter("response_type", "code")
-            .appendQueryParameter("scope", "openid email phone offline_access")
+            .appendQueryParameter("scope", "openid offline_access")
```
elishkaflint, line 15: My bad - but the ticket was incorrect - the scope should just be `openid`

We corrected this when we did the integration between iOS and the stub but I forgot to go back and update the Android tickets. 

That's done now, please can we update this accordingly? Thank you and sorry for catching this so late!

elishkaflint: Screenshot 2023-11-08 at 16 14 21

app/src/staging/res/values/login_flow.xml
elishkaflint: Confirmed that this is not considered a secret here https://gds.slack.com/archives/C04L0BZQNR3/p1699459522702629

Please feel free to resolve this convo once you've seen it and are happy @MohamedHassan101

app/src/main/java/uk/gov/onelogin/login/UriBuilder.kt
MohamedHassan101: no problem , will update that now

app/src/staging/res/values/login_flow.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="openIdConnectBaseUrl" translatable="false">https://oidc.staging.account.gov.uk%1$s</string>
-    <string name="openIdConnectClientId" translatable="false">CLIENT_ID</string>
-    <string name="webBaseUrl" translatable="false">https://mobile-staging.account.gov.uk%1$s</string>
-</resources>
\ No newline at end of file
+    <string name="openIdConnectBaseUrl" translatable="false">https://oidc.integration.account.gov.uk%1$s</string>
```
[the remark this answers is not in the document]
MohamedHassan101: This is used to appending one string to the end of another string.In this case we are adding "/authorize" at the end of https://oidc.integration.account.gov.uk%1$s

app/src/build/res/values/login_flow.xml
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="openIdConnectBaseUrl" translatable="false">https://oidc.build.account.gov.uk%1$s</string>
-    <string name="openIdConnectClientId" translatable="false">CLIENT_ID</string>
+    <string name="openIdConnectBaseUrl" translatable="false">https://auth-stub.mobile.dev.account.gov.uk%1$s</string>
+    <string name="openIdConnectClientId" translatable="false">TEST_CLIENT_ID</string>
     <string name="webBaseUrl" translatable="false">https://mobile-build.account.gov.uk%1$s</string>
+
```
elishkaflint, line 6: can we remove these blank lines?

app/src/staging/res/values/login_flow.xml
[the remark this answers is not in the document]
cgedrim-co: It's an `printf`-ish placeholder, so no crazy encoding or anything 😉 

You'd use it like this:

```
<string name="example">Hello %1$s, good %2$s</string>
```

```
resources.getString(
  R.string.example,
  "Elishka",
  "evening"
)
```

Which would render "Hello Elishka, good evening"

elishkaflint (approved):

PR #55, opened 2023-11-27
DCMAW-7472: Implement Hilt and write E2E tests

cgedrim-co:
- Adds annotation for Hilt to allow for injecting mocks into our end-to-end tests

Resolves: DCMAW-7472

_EDIT (by @alex-bradbury )_
The E2E test is `ignored` currently as I have been unable to get the test to work on the pipeline emulator. This will be fixed in a follow up ticket


app/src/main/java/uk/gov/onelogin/login/nonce/NonceGeneratorModule.kt
```diff
@@ -0,0 +1,16 @@
...
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+object NonceGeneratorModule {
+
+    @Provides
+    @Singleton
+    fun bindsNonceGenerator(nonceGenerator: NonceGenerator): INonceGenerator = nonceGenerator
+}
```
Chuddington, line 16: If the implementations are injectable, as shown by having access to the `NonceGenerator` in the function parameters, we're able to change this object into an interface, along with the value. 

Which style do you prefer with your hilt modules?

```suggestion
@Module
@InstallIn(SingletonComponent::class)
interface NonceGeneratorModule {

    @Binds
    @Singleton
    fun bindsNonceGenerator(nonceGenerator: NonceGenerator): INonceGenerator
}
```

app/src/main/java/uk/gov/onelogin/login/state/IStateGenerator.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.onelogin.login.state
+
+interface IStateGenerator {
+    fun generate(): String
+}
```
Chuddington, line 5: As this only has one function, we're able to convert this into a functional interface. This has the real nice benefit of how you can go about creating anonymous implementations - `val something = IStateGenerator { "returnSomeString" }`.

We can then declare test implementations like so:

```suggestion
fun interface IStateGenerator {
    fun generate(): String
    
    companion object {
    
        /**
         * Create an [IStateGenerator] instance that returns the provided [state]
         * for the [IStateGenerator.generate] function.
         *
         * Use within tests for providing custom data.
         *
         * @return The [state] parameter.
         */
        @VisibleForTesting(otherwise = VisibleForTesting.private)
        fun stub(state: String) = IStateGenerator { state }
        
        /**
         * Create an [IStateGenerator] instance that returns the [state.invoke] value
         * for the [IStateGenerator.generate] function.
         *
         * Use within tests for providing custom data.
         *
         * @return The [state.invoke] return value.
         */
        @VisibleForTesting(otherwise = VisibleForTesting.private)
        fun stub(state: () -> String) = IStateGenerator { state.invoke() }
        
        /**
         * Create an [IStateGenerator] instance that causes an [Exception] upon use.
         *
         * Use within tests that don't require [IStateGenerator] usage whilst still needing
         * the presence of a generator as a way of marking those tests as unrelated.
         */
        @VisibleForTesting(otherwise = VisibleForTesting.private)
        fun dummy() = IStateGenerator {
            throw IllegalAccessException(
                "This IStateGenerator instance should not have been used!"
            )
        }
    }
}
```

Do you think there's value in having the preceding companion object? If so, do we want to do this with other interfaces that have a single function within them?  I like that there's only one function in things

app/src/main/java/uk/gov/onelogin/login/state/StateGenerator.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.onelogin.login.state
+
+import java.util.UUID
+import javax.inject.Inject
+
+class StateGenerator @Inject constructor() : IStateGenerator {
```
Chuddington, line 6: How do you feel about the name `UUIDStateGenerator`? I feel like we've got the power to name it to have a little more context. Especially with it being a little class, yeah.

authentication/build.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+            )
+        }
+    }
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
+    }
+    kotlinOptions {
+        jvmTarget = "1.8"
+    }
```
Chuddington, line 33: Does your `buildLogic` project have a plugin for it that relates to setting up java configuration? If there is, I think you're able to configure / use that, then this block can be removed.

However, I don't know whether you would have an 'android library' version of the plugin in there. I believe the wallet repository would have that version as I remember talking it through with @chrisdondig .

authentication/src/main/java/uk/gov/android/authentication/ILoginSession.kt
```diff
@@ -0,0 +1,14 @@
...
+import android.content.Intent
+import net.openid.appauth.TokenResponse
+
+interface ILoginSession {
+    fun present(
+        configuration: LoginSessionConfiguration
+    )
+
+    fun init(context: Context): ILoginSession
+    fun finalise(intent: Intent, callback: (tokens: TokenResponse) -> Unit)
```
Chuddington, line 13: I dunno how I feel about exposing these third party dependencies as function parameters.

If that's the application or activity context in `init()`, we can make that a Hilt injectable thing within an implementation's constructor so passing the context isn't required.

authentication/src/main/java/uk/gov/android/authentication/LoginSession.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    override fun init(
+        context: Context
+    ): ILoginSession {
+        if (this.context == null) {
+            this.context = context
+            authService = AuthorizationService(context)
+        }
+        return this
+    }
```
Chuddington, line 27: This sort of goes into one of the preceding comments - if it's worth using an Injectable context over the use of this function and `AuthorizationService` instantiation?

authentication/src/main/java/uk/gov/android/authentication/LoginSession.kt
```diff
@@ -0,0 +1,100 @@
...
+
+            val authRequest = builder.build()
+
+            val authIntent = authService.getAuthorizationRequestIntent(authRequest)
+            ActivityCompat.startActivityForResult(
+                context as Activity,
+                authIntent,
+                REQUEST_CODE_AUTH,
+                null
+            )
```
Chuddington, line 71: What do we use `ActivityCompat.startActivityForResult` for? I ask because I'm wondering what would happen if the Activity state isn't being monitored somehow: https://developer.android.com/training/basics/intents/result#register

Chuddington: Ah it's the `@ActivityContext` that's being used: https://developer.android.com/training/dependency-injection/hilt-android#predefined-qualifiers

build.gradle.kts
```diff
@@ -9,55 +8,51 @@ plugins {
...
+
+fun getVersionCode(): Int {
+    val code: Int = if (rootProject.hasProperty("versionCode")) {
+        rootProject.property("versionCode").toString().toInt()
+    } else if (localProperties.getProperty("versionCode") != null) {
+        localProperties.getProperty("versionCode").toString().toInt()
+    } else {
+        throw Error(
+            "Version code was not found as a command line parameter or a local property"
+        )
```
Chuddington, line 32: Do we have any documentation somewhere around the potential next steps when encountering this error?

gradle/libs.versions.toml
```diff
@@ -3,35 +3,100 @@
...
 [versions]
 
 kotlin = "1.9.21"
-##  ⬆ = "2.0.0-Beta1"
 
 androidx-databinding = "8.1.3"
+agp = "8.2.0"
+core-ktx = "1.9.0"
+##    ⬆ = "1.10.0-alpha01"
+##    ⬆ = "1.10.0-alpha02"
```
Chuddington, line 17: Yo, is this the `refreshVersions` plugin doing some magic? This is an exciting prospect if it is

versions.properties
Chuddington, line 1: We've got a properties file as well...

Do you know whether it's possible to remove this, if we've got versions being provided in the gradle version catalog?

authentication/src/main/java/uk/gov/android/authentication/ILoginSession.kt
alex-bradbury: This has been changed and refactored as you have suggested in the actual seperate repo here: https://github.com/govuk-one-login/mobile-android-authentication/blob/main/app/src/main/java/uk/gov/android/authentication/LoginSession.kt

This code will be removed from this repo soon

authentication/src/main/java/uk/gov/android/authentication/LoginSession.kt
alex-bradbury: See above reply

versions.properties
alex-bradbury: See Jira ticket

build.gradle.kts
alex-bradbury: I'm not sure we do, something to look into I suppose

authentication/build.gradle.kts
alex-bradbury: In jira ticket

app/src/main/java/uk/gov/onelogin/login/state/StateGenerator.kt
alex-bradbury: In jira ticket

app/src/main/java/uk/gov/onelogin/login/state/IStateGenerator.kt
alex-bradbury: In jira ticket

app/src/main/java/uk/gov/onelogin/login/nonce/NonceGeneratorModule.kt
alex-bradbury: In jira ticket

app/src/main/java/uk/gov/onelogin/home/HomeScreen.kt
```diff
@@ -50,29 +51,32 @@ fun HomeScreen(
                     fontWeight = FontWeight.Bold
                 )
                 Text(
-                    tokens.access,
+                    tokens?.accessToken ?: "No access token set!",
                     modifier = Modifier
                         .padding(16.dp)
```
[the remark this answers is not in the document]
alex-bradbury: In jira ticket

app/src/e2eTestBuild/java/uk/gov/onelogin/login/SuccessfulLoginTest.kt
```diff
@@ -0,0 +1,223 @@
...
+
+    private fun setupNavigation() {
+        navController = composeTestRule.setupComposeTestRule { _ ->
+            LoadingScreen()
+        }
+    }
+
+    @Ignore("This test is failing on the pipeline emulator, we're struggling to figure out why")
+    @Test
+    fun logsIntoTheAppAndExchangesAuthCodeForTokens() {
```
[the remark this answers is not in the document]
alex-bradbury: All comments on this class, we have a follow up ticket to fix this test and get it working, I would appreciate for now keeping the work in progress effort here. Not ideal I know but we need to move forward with this

app/src/e2eTestBuild/java/uk/gov/onelogin/test/settings/SettingsController.kt
```diff
@@ -0,0 +1,288 @@
...
+    private fun openSettings3() {
+        device.pressHome()
+        val drawerIcon = device.findObject(UiSelector().description("Apps"))
+        drawerIcon.clickAndWaitForNewWindow()
+        val appToTest =
+            device.findObject(UiSelector().description("Settings"))
+        appToTest.clickAndWaitForNewWindow()
+    }
+
+    private fun openSettings4() {
```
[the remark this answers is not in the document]
alex-bradbury: see above comment

app/src/e2eTestBuild/java/uk/gov/onelogin/test/settings/SettingsController.kt
```diff
@@ -0,0 +1,288 @@
...
+            UiSelector().text("Open by default")
+        )
+
+        try {
+            settingsPage.scrollIntoView(openByDefaultSelector)
+        } catch (e: UiObjectNotFoundException) {
+            e.printStackTrace()
+        }
+
+//        takeScreenshot("screenshot2.png")
```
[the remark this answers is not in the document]
alex-bradbury: see above comment

app/src/e2eTestBuild/java/uk/gov/onelogin/test/settings/SettingsController.kt
```diff
@@ -0,0 +1,288 @@
...
+            UiSelector().text("Open supported links")
+        )
+
+        try {
+            openByDefaultPage.scrollIntoView(openSupportedLinksSelector)
+        } catch (e: UiObjectNotFoundException) {
+            e.printStackTrace()
+        }
+
+//        takeScreenshot("screenshot3.png")
```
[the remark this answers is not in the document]
alex-bradbury: see above comment

Chuddington (approved):
You've captured the changes in a ticket somewhere so hopefully we'll get the opportunity to work with it

geoffreyssh97 (approved):

PR #71, opened 2024-02-08
DCMAW-7798: Implement 'passcode enabled' check

alex-bradbury:
### Fix for STS authorise call
- add sts client_id
- add sts scope

Resolves: DCMAW-7624

### Add Passcode Check after login
- Create `CredentialChecker` interface and `DeviceCredentialChecker` implementation to check if phone has passcode/pin
- Add placeholder error screen if device is not secure
- refactor `MainActivityViewModel` to remove need for context within

Resolves: DCMAW-7798

### DIFF
--- a/app/src/main/java/uk/gov/onelogin/MainActivityViewModel.kt
+++ b/app/src/main/java/uk/gov/onelogin/MainActivityViewModel.kt
@@ -1,23 +1,23 @@
 package uk.gov.onelogin
 
-import android.content.Context
 import android.content.Intent
+import android.content.SharedPreferences
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
 import dagger.hilt.android.lifecycle.HiltViewModel
 import uk.gov.android.authentication.LoginSession
 import uk.gov.android.authentication.TokenResponse
+import uk.gov.onelogin.credentialchecker.CredentialChecker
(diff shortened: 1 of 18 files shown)


app/src/main/java/uk/gov/onelogin/MainActivityViewModel.kt
```diff
@@ -26,55 +26,52 @@ class MainActivityViewModel @Inject constructor(
...
         intent: Intent?,
-        context: Context
+        sharedPrefs: SharedPreferences
     ) {
         val data = intent?.data
         when {
-            data != null -> handleActivityResult(context, intent)
-            tokensAvailable(context) -> _next.value = HomeRoutes.START
+            data != null -> handleActivityResult(sharedPrefs, intent)
+            tokensAvailable(sharedPrefs) -> _next.value = HomeRoutes.START
```
flavius-mester, line 34: Do we need to check if `credChecker.isDeviceSecure()` when they resume to the app and before using the tokens again?

flavius-mester (approved):
Code looks good to me! 🚀

app/src/main/java/uk/gov/onelogin/MainActivityViewModel.kt
alex-bradbury: It's a good point and possibly a question for product. Because this is part of the initial first time user setup check what level of authentication we can ask for.
Will raise with Dylan

PR #79, opened 2024-02-20
Add Generic Error Screen & Route

flavius-mester:
Add Generic Error Screen
Add Route that can be reused from anywhere in the graph
Add utility extension so the generic error screen can be reached easier
Use the generic error screen when Login Fails

**Testing:**
Login, Click on the error scenarios from the stub environment

| Light Mode      | Dark Mode       |
| -------------- | -------------- |
| Screenshot_20240221_081931    |   Screenshot_20240221_081853   |


app/src/main/java/uk/gov/onelogin/ui/error/ErrorRoutes.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import androidx.navigation.NavController
+import androidx.navigation.NavGraphBuilder
+import androidx.navigation.NavHostController
+import androidx.navigation.compose.composable
+import androidx.navigation.navigation
+
+object ErrorRoutes {
+    const val ROOT: String = "/error"
+    const val START: String = "$ROOT/start"
```
alex-bradbury, line 11: Maybe this should be called `GENERIC`?  not sure start makes much sense here

alex-bradbury (commented):
Looks great!
One small comment there

app/src/main/java/uk/gov/onelogin/ui/error/ErrorRoutes.kt
flavius-mester: yeah, `generic` makes more sense here  than `start`. I had it generic before, not sure why I renamed it back to start. Maybe to keep it consistent with the rest of the routes. [In order to be able to launch the generic screen from anywhere in the graph, it needs to be mounted in as a new sub graph route, so we have to have a `start` destination. Although it might be the only one

flavius-mester: done [98f13d0](https://github.com/govuk-one-login/mobile-android-one-login-app/pull/79/commits/98f13d05cf61d0c1725bdc1f3682eb6cc4b570ff)

alex-bradbury (approved):

PR #80, opened 2024-02-20
DCMAW-7847: biometrics check

alex-bradbury:
- add function to `CredentialChecker` to allow for checking the biometric status of the device
- implement this function in the `DeviceCredentialChecker`
- create our own `BiometricManager`, this simply maps the AndroidX manager to return values that we own
- Add routing login into `MainActivityViewModel`


app/src/main/java/uk/gov/onelogin/login/BiometricsOptInScreen.kt
```diff
@@ -0,0 +1,41 @@
...
+        Column(
+            horizontalAlignment = Alignment.CenterHorizontally,
+            verticalArrangement = Arrangement.Center,
+            modifier = Modifier
+                .padding(16.dp)
+                .fillMaxSize()
+                .verticalScroll(rememberScrollState())
+        ) {
+            Text(
+                text = "Placeholder",
```
flavius-mester, line 30: Placeholder?

flavius-mester (approved):

app/src/main/java/uk/gov/onelogin/login/BiometricsOptInScreen.kt
alex-bradbury: Ah this is because the actual screen will be built in the next ticket, so for now the screen is purely a placeholder to allow the routing to happen properly

PR #82, opened 2024-02-26
Feat/DCMAW-7849: bio opt in screen

alex-bradbury:
- Add necessary strings in english and welsh
- build biometrics opt-in screen ui with vector image
- Move screens into ui package
- Designs for Opt In screen
- Create BiometricPreferenceHandler and implementation for saving user authentication preference
- Set pref when device is not secured in MainActivityViewModel


flavius-mester (approved):
Looking great 🎸

PR #83, opened 2024-03-07
DCMAW-7543: Add Sonar gate quality check to PR workflow

geoffreyssh97:
https://govukverify.atlassian.net/browse/DCMAW-7543

This PR adds a conditional step in the Sonar action file, to check the result of the quality gate scan.


flavius-mester (approved):
Do you want to do this only on `on_pull_request`? how about `on_push(develop)`

geoffreyssh97:
> Do you want to do this only on `on_pull_request`? how about `on_push(develop)`

Currently we are only making this change for PR workflows, to keep it consistent with other repos. For some repos, there are Sonar errors on the `main` branch so adding the Quality Gate check to fail would cause 'on push' workflows to fail. When these errors are resolved, we may then amend post-merge workflows to also run the Quality Gate check.

PR #90, opened 2024-03-25
feat/DCMAW-7898: Save Tokens to Secure Store

alex-bradbury:
- Create TokenRepository to handle saving the tokens in working memory
- Create SaveToSecureStore use case to handle saving to SecureStore instance and centralise the creation of the SharedPrefsStore instance. Ensures creation only happens when a BioPref has been set
- Create SaveTokenExpiry use case to handle saving the token expiry in shared prefs
- Keys object created to house keys related to the tokens being saved in shared prefs and secure storage etc.
- Mapper method handles mapping bio prefs to access control level for secure storage
- Change in MainActivityViewModel to stop saving the tokens in shared prefs and into the new repository instead
- HomeScreen now handles saving the tokens into SecureStore via methods in HomeScreenViewModel
- Add relevant tests

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -161,6 +161,7 @@ dependencies {
         libs.pages,
         libs.slf4j.api,
         libs.theme,
+        libs.secure.store,
         libs.authentication,
         projects.features
     ).forEach(::implementation)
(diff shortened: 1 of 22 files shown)


flavius-mester (approved):
Looking good 🚀

PR #101, opened 2024-04-25
feat/DCMAW-6346: nav tabs

alex-bradbury:
### Create Compose screen for main bottom navigation
- remove local `GdsNavigationBar` and associated classes, now using from UI module
- add icons for home, wallet, profile
- create nav destinations for bottom tabs in `BottomNavDestination`
- Fix bio opt in bug

### Add Tab navigation to home screen
- Refactor `HomeRoutes` to be `MainNavRoutes`, this now handles the tab navigation scaffold
- `BottomNavDestination` created to handle the tabs required
- Various files moved around to make more package structure sense
- Navigation in `MainActivity` set to close screen once moved from it. Now users can't go 'back' to signin and splash screens etc.
- Check if bio pref in `HandleLogin` is not `NONE`, if so we need to do full login (bug fix)


app/src/main/java/uk/gov/onelogin/MainActivity.kt
```diff
@@ -36,7 +36,11 @@ class MainActivity : AppCompatActivity() {
             LaunchedEffect(key1 = Unit) {
                 viewModel.apply {
                     next.observe(lifecycleOwner) {
-                        navController.navigate(it)
+                        navController.navigate(it) {
+                            popUpTo(navController.graph.id) {
```
flavius-mester, line 40: what does this do exactly? 🤔

app/src/main/java/uk/gov/onelogin/mainnav/nav/MainNavRoutes.kt
```diff
@@ -1,22 +1,23 @@
...
+    fun NavGraphBuilder.mainNavRoutesFlow() {
         navigation(
             route = ROOT,
             startDestination = START
         ) {
             composable(
                 route = START
             ) {
-                HomeScreen()
+                MainNavScreen()
```
flavius-mester, line 20: You don't want to pass the NavController to the `MainNavScreen`?

app/src/main/java/uk/gov/onelogin/mainnav/ui/MainNavScreen.kt
```diff
@@ -0,0 +1,105 @@
...
+import uk.gov.onelogin.mainnav.nav.BottomNavDestination
+import uk.gov.onelogin.ui.components.SimpleTextPage
+import uk.gov.onelogin.ui.home.HomeScreen
+import uk.gov.ui.components.navigation.GdsNavigationBar
+import uk.gov.ui.components.navigation.GdsNavigationItem
+
+@Suppress("LongMethod")
+@Composable
+fun MainNavScreen(
+    navController: NavHostController = rememberNavController()
```
flavius-mester, line 36: Do you want a new navController?
Can this new navController reach some destinations defined in the main one. Like generic error screens? etc

app/src/main/java/uk/gov/onelogin/MainActivity.kt
alex-bradbury: essentially clears the back stack. So this navigation is always off either splash screen or sign in screen so think we want to clear them out the back stack

flavius-mester: and we want to do this every single time we handle a navigation?

It might work for now, but let's keep an eye on it.

app/src/main/java/uk/gov/onelogin/mainnav/nav/MainNavRoutes.kt
alex-bradbury: We need to look into this, how we do the nested navigation, think something to come back for another day

app/src/main/java/uk/gov/onelogin/mainnav/ui/MainNavScreen.kt
alex-bradbury: see above comment about this

flavius-mester (approved):

PR #104, opened 2024-05-08
fix/refactor

alex-bradbury:
- Remove old local network package stuff and import android-networking module as dependency
- fix Resource linking
- Remove nonce and AppRoutes


flavius-mester (approved):
Great tidy up! we should do more of this 💯

PR #105, opened 2024-05-09
feat/DCMAW-8633: auth provider implementation

alex-bradbury:
- Created `StsAuthenticationProvider`, takes a tokenRepo and httpClient to allow to get the access token then make a call to sts to exchange for service token
- `TokenResponse` created to model service token response


flavius-mester (approved):
looks good!

PR #107, opened 2024-05-14
feat/DCMAW-8814: Create HelloWorld sts call

alex-bradbury:
- Creation of `AuthTabScreen` to live in the developer menu. This has two buttons to allow a happy and sad path call to `/hello-world` endpoint
- `AuthTabScreenViewModel` handles making the calls to hello world. Sad call temporarily deletes the stored access token to simulate an error
- STS endpoint feature is turned on
- Use case `HelloWorldApiCall` created to handle calls to `/hello-world` endpoint
- Necessary hilt modules created to provide http clients from the android-network module
- Slight changes in `WelcomeScreenViewModel` to fix calling the STS `/authorize` url

Resolves: DCMAW-8814

### DIFF
--- a/app/src/main/java/uk/gov/onelogin/developer/tabs/auth/AuthTabScreenViewModel.kt
+++ b/app/src/main/java/uk/gov/onelogin/developer/tabs/auth/AuthTabScreenViewModel.kt
@@ -1,0 +1,54 @@
+package uk.gov.onelogin.developer.tabs.auth
+
+import androidx.compose.runtime.State
+import androidx.compose.runtime.mutableStateOf
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+import kotlinx.coroutines.launch
+import uk.gov.onelogin.network.usecase.HelloWorldApiCall
+import uk.gov.onelogin.repositiories.TokenRepository
+
+@HiltViewModel
+class AuthTabScreenViewModel @Inject constructor(
+    private val helloWorldApiCall: HelloWorldApiCall,
+    private val tokenRepository: TokenRepository
+) : ViewModel() {
+    private val _happyHelloWorldResponse = mutableStateOf("")
+    val happyHelloWorldResponse: State<String>
+        get() = _happyHelloWorldResponse
+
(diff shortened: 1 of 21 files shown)


app/src/main/java/uk/gov/onelogin/developer/tabs/auth/AuthTabScreenViewModel.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    fun makeHappyHelloWorldCall() {
+        _happyCallLoading.value = true
+        viewModelScope.launch {
+            _happyHelloWorldResponse.value = helloWorldApiCall()
+            _happyCallLoading.value = false
+        }
+    }
+
+    fun makeSadHelloWorldCall() {
```
flavius-mester, line 42: This works!

What I was having in mind was to perhaps modify the other Tab. `SecureStor` and add a button there `Purge token from Runtime`, `Restore token to Runtime`, `Purge token from Disk` etc.

Just to the QA(and us/other devs) has visibility of why the network request fails.

flavius-mester (commented):
Looking good, left some comments

app/src/main/java/uk/gov/onelogin/developer/tabs/auth/AuthTabScreenViewModel.kt
alex-bradbury: mmmm true, think that functionality would be nice but maybe too much to add to this 'ticket'

app/src/main/java/uk/gov/onelogin/network/di/NetworkModule.kt
```diff
@@ -30,15 +33,34 @@ object NetworkModule {
...
+        val appName = context.resources.getString(R.string.app_name)
+        userAgentGenerator.setUserAgent(
+            UserAgent(
+                appName = appName,
+                versionName = BuildConfig.VERSION_NAME,
+                clientName = KTOR_CLIENT,
+                manufacturer = Build.MANUFACTURER,
+                model = Build.MODEL,
+                sdkVersion = Build.VERSION.SDK_INT,
+                clientVersion = BuildConfig.KTOR_VERSION
```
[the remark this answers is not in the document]
alex-bradbury: yeah good point, looked at what Wallet were doing but actually very much agree, will change to app version

PR #108, opened 2024-05-15
Add Titled Screen to Home, Wallet and Profile screen

flavius-mester:
Add Titled page that is scrollable
https://govukverify.atlassian.net/browse/DCMAW-8778

Screenshot_20240515_183120
Screenshot_20240515_183133
Screenshot_20240515_183140


app/src/main/java/uk/gov/onelogin/ui/profile/ProfileScreen.kt
```diff
@@ -0,0 +1,47 @@
...
+import uk.gov.android.ui.pages.TitledPage
+import uk.gov.android.ui.pages.TitledPageParameters
+
+@Composable
+fun ProfileScreen() {
+    TitledPage(
+        parameters = TitledPageParameters(
+            R.string.app_profile
+        ) {
+            HorizontalDivider()
```
alex-bradbury, line 26: Could probably build this email display section into a separate composable given it's used twice...

flavius-mester: That's what I was thinking as well, but then I realised it's mostly one component? (a formatter Text field, surrounded by dividers).
And I've also struggled to come up with a name that is not `strategic app specific`, so it would make sense to have it in the Design System Library. `TwoLineHeader`?

alex-bradbury: I don't think it needs to be in the UI repo, I think you can call it some thing like `EmailHeader` and keep it somewhere in this repo

flavius-mester: yeah, that makes sense. We need `core/ui`

flavius-mester: done d10d7669d24ddcb7138b0ba32ae189aec6979ade

PR #124, opened 2024-05-21
Create Feature Flag Selection Tab

alex-bradbury:
- Composable created for feature selection screen
- Screen added to developer panel
- Viewmodel created to handle logic
- `AvailableFeatures` created to handle management of available features

Resolves: DCMAW-8268

### DIFF
--- a/app/src/main/java/uk/gov/onelogin/login/ui/welcome/WelcomeScreenViewModel.kt
+++ b/app/src/main/java/uk/gov/onelogin/login/ui/welcome/WelcomeScreenViewModel.kt
@@ -53,11 +53,8 @@ class WelcomeScreenViewModel @Inject constructor(
             context.resources.getString(R.string.openIdConnectClientId)
         }
 
-        val scopes = if (featureFlags[StsFeatureFlag.STS_ENDPOINT]) {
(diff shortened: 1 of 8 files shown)


app/src/main/java/uk/gov/onelogin/login/ui/welcome/WelcomeScreenViewModel.kt
```diff
@@ -53,11 +53,8 @@ class WelcomeScreenViewModel @Inject constructor(
             context.resources.getString(R.string.openIdConnectClientId)
         }
 
-        val scopes = if (featureFlags[StsFeatureFlag.STS_ENDPOINT]) {
-            listOf(LoginSessionConfiguration.Scope.OPENID)
-        } else {
-            listOf(LoginSessionConfiguration.Scope.STS)
-        }
+        val scopes = listOf(LoginSessionConfiguration.Scope.OPENID)
```
flavius-mester, line 56: So we don't use STS anymore?

alex-bradbury: this seems to be the only way to get the call to auth to work

flavius-mester (approved):

PR #126, opened 2024-05-22
Add bundler setup code for linux

flavius-mester:
Fixes https://govukverify.atlassian.net/browse/DCMAW-9212


config/actions/setup-runner/action.yml
```diff
@@ -27,12 +27,25 @@ runs:
...
+    - name: Install Gemfile dependencies
       run: |
-        sudo snap install vale --edge
+        export LC_ALL=en_GB.UTF-8
+        export LANG=en_GB.UTF-8
+        export GEM_HOME=$HOME/.gem
+        export GEM_PATH=$HOME/.gem
+        eval "$(rbenv init - bash)"
+
+        rbenv install --skip-existing
```
Chuddington, line 46: `rbenv` won't know what version of ruby to run so you need a `.ruby-version` file so it can.

This file just contains the version number of ruby you're looking for. Within the ID Check space, this is just `3.3.0`.

Brewfile
```diff
@@ -0,0 +1,14 @@
...
+# `brew bundle --no-lock` in the root directory of the repository.
+
+# Taps (data sources)
+tap "homebrew/cask-versions" # Access alternative versions for homebrew casks
+
+# terminal programs
+brew "rbenv" unless system "rbenv", "--version"
+brew "rbenv-bundler" unless system "which", "rbenv-bundler"
+brew "ruby-build" unless system "ruby-build", "--version"
+brew "vale" unless system "vale", "--version"
```
Chuddington, line 11: The Brewfile should also facilitate developers so that there's one set of dependencies required for the code base. A big example would be onboarding new joiners, right?. With this in mind, what would be the reason you've avoided bringing in tooling such as GPG (for developer commits) and AWS (for secrets management)?

Chuddington (changes requested):

Brewfile
flavius-mester: My goal was just to get our pipeline working again. It's a good idea to streamline and standardise all our tools/versions among devs though 👍

config/actions/setup-runner/action.yml
flavius-mester: Originally I tried installing `3.3.0` with apt, and struggled as the version was not available in the registry. The reason why I am trying to move to `brew` is to circumvent that issue

flavius-mester:
Dropped this PR in favour of this -> https://github.com/govuk-one-login/mobile-android-one-login-app/pull/128#issue-2313555622

PR #129, opened 2024-05-24
Tidying Versions

alex-bradbury:
- versions.properties removed
- convert all versions in relevant build.gradle files to use libs
- remove splash screen stuff


flavius-mester (approved):
Plenty of changes, hope it all builds well once in develop as well :D

PR #139, opened 2024-06-06
feat/DCMAW-8574: verify id token

alex-bradbury:
## Add jose4j verifier
- Create JwtVerifier interface and jose4j implementation
- JwkDeserializer to handle key deserialization

## Create usecase for verifying token
- usecase makes called to jwks endpoint, then uses the key to verify a given token using the `JwtVerifier`

## Verify and save Id token
- HomeScreenViewModel now saves access and id tokens as separate secure store entries
- VerifyIdToken use cases managed making the call to `/jwks` endpoint and using the `JwtVerifier` to use the provided key to verify the id token
- MainActivityViewModel not uses the verify id token usecase as part of the login process


app/src/main/java/uk/gov/onelogin/MainActivityViewModel.kt
```diff
@@ -62,6 +60,38 @@ class MainActivityViewModel @Inject constructor(
...
+
+        tokens.idToken?.let { idToken ->
+            verifyIdToken(idToken, jwksUrl) { verified ->
+                if (!verified) {
+                    _next.value = LoginRoutes.SIGN_IN_ERROR
+                } else {
+                    checkLocalAuthRoute(tokens)
+                }
+            }
+        } ?: checkLocalAuthRoute(tokens)
```
flavius-mester, line 77: if the idToken is missing we carry on with login?

app/src/main/java/uk/gov/onelogin/login/usecase/VerifyIdToken.kt
```diff
@@ -0,0 +1,86 @@
...
+class VerifyIdTokenImpl @Inject constructor(
+    private val httpClient: GenericHttpClient,
+    private val verifier: JwtVerifier
+) : VerifyIdToken {
+    override fun invoke(
+        idToken: String,
+        jwksUrl: String,
+        callback: (Boolean) -> Unit
+    ) {
+        MainScope().launch {
```
flavius-mester, line 35: shall we use `IO` for this? or the networking client does this automatically?

flavius-mester (approved):
Strong PR. Looks good.

Please check he comment and feel free to merge if you are comfortable with the `idToken` being null

app/src/main/java/uk/gov/onelogin/login/usecase/VerifyIdToken.kt
alex-bradbury: yep, refactored this bit, shouldn't really be having a global scope being launched

app/src/main/java/uk/gov/onelogin/MainActivityViewModel.kt
flavius-mester: hmm, yeah alright. Maybe in future we will need to optionally disable the verify in lower environments like `dev`? or STS will be available in dev as well?

PR #147, opened 2024-06-13
Add email to home

flavius-mester:
https://govukverify.atlassian.net/browse/DCMAW-8650

Screenshot 2024-06-17 at 12 17 14

### DIFF
--- a/app/src/main/java/uk/gov/onelogin/ui/home/HomeScreen.kt
+++ b/app/src/main/java/uk/gov/onelogin/ui/home/HomeScreen.kt
@@ -27,11 +27,12 @@ fun HomeScreen(
 ) {
     viewModel.saveTokens(LocalContext.current as FragmentActivity)
     val tokens = viewModel.getTokens()
+    val email = viewModel.getEmail() ?: ""
     TitledPage(
         parameters = TitledPageParameters(
             R.string.app_homeTitle
         ) {
-            EmailHeader("someEmail@mail.com")
+            EmailHeader(email)
             Text(
                 text = "Access Token",
                 fontWeight = FontWeight.Bold,
--- a/app/src/main/java/uk/gov/onelogin/ui/home/HomeScreen.kt
+++ b/app/src/main/java/uk/gov/onelogin/ui/home/HomeScreen.kt
@@ -27,7 +27,7 @@ fun HomeScreen(
 ) {
     viewModel.saveTokens(LocalContext.current as FragmentActivity)
     val tokens = viewModel.getTokens()
-    val email = viewModel.getEmail() ?: ""
+    val email = viewModel.email
     TitledPage(
         parameters = TitledPageParameters(
             R.string.app_homeTitle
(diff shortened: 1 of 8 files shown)


app/src/main/java/uk/gov/onelogin/ui/home/HomeScreenViewModel.kt
```diff
@@ -9,14 +9,16 @@ import kotlinx.coroutines.launch
...
 import uk.gov.onelogin.tokens.usecases.SaveToSecureStore
 import uk.gov.onelogin.tokens.usecases.SaveTokenExpiry
 
 @HiltViewModel
 class HomeScreenViewModel @Inject constructor(
     private val tokenRepository: TokenRepository,
     private val saveToSecureStore: SaveToSecureStore,
-    private val saveTokenExpiry: SaveTokenExpiry
+    private val saveTokenExpiry: SaveTokenExpiry,
+    val getEmail: GetEmail
```
alex-bradbury, line 21: This should be private surely?

Then we should probably have some `LiveData` to handle the email in the ui

app/src/main/java/uk/gov/onelogin/ui/home/HomeScreen.kt
```diff
@@ -27,11 +27,12 @@ fun HomeScreen(
 ) {
     viewModel.saveTokens(LocalContext.current as FragmentActivity)
     val tokens = viewModel.getTokens()
+    val email = viewModel.getEmail() ?: ""
```
alex-bradbury, line 30: See below, think this should be `LiveData`

app/src/test/java/uk/gov/onelogin/tokens/usecases/GetEmailImplTest.kt
```diff
@@ -0,0 +1,33 @@
...
+    fun setUp() {
+        whenever(tokenResponse.idToken).thenReturn(idToken)
+        whenever(tokenRepository.getTokenResponse()).thenReturn(tokenResponse)
+    }
+
+    @Test
+    operator fun invoke() {
+        val emailResponse = sut.invoke()
+        assertEquals(expectedEmail, emailResponse)
+    }
```
alex-bradbury, line 32: Can we have some tests in here to check null return results if, for example, the id token is null or the id token doesn't contain an email field etc.

flavius-mester: good call -> 50d2dcd248ea1121c47ddb635f7ab7d363638768

app/src/main/java/uk/gov/onelogin/ui/home/HomeScreenViewModel.kt
flavius-mester: as we discussed -> 1d72c76af5b623681ee7d7cd31d46f289a480372

app/src/main/java/uk/gov/onelogin/ui/home/HomeScreen.kt
flavius-mester: same 1d72c76af5b623681ee7d7cd31d46f289a480372

alex-bradbury (approved):

PR #169, opened 2024-07-05
Sign out error screen

flavius-mester:
https://govukverify.atlassian.net/browse/DCMAW-9096

Screenshot 2024-07-11 at 10 23 54

### DIFF
--- a/app/src/main/java/uk/gov/onelogin/signOut/ui/SignOutScreen.kt
+++ b/app/src/main/java/uk/gov/onelogin/signOut/ui/SignOutScreen.kt
@@ -17,6 +17,7 @@ import uk.gov.android.ui.pages.AlertPageParameters
 import uk.gov.onelogin.signOut.domain.SignOutError
 
 @Composable
+@Suppress("SwallowedException")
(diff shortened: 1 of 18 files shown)


app/src/main/java/uk/gov/onelogin/signOut/ui/SignOutScreen.kt
```diff
@@ -40,8 +42,12 @@ fun SignOutScreen(
                 goBack()
             },
             onPrimary = {
-                viewModel.signOut(context)
-                goToSignIn()
+                try {
+                    viewModel.signOut(context)
+                    goToSignIn()
+                } catch (e: SignOutError) {
+                    goToSignOutError()
```
alex-bradbury, line 49: We can remove the suppression of the swallowed exception if we just log the error here

flavius-mester: done 7371b4188da43dbf62772c5fc06dabb6e51b25ca

alex-bradbury (approved):

PR #177, opened 2024-07-31
feat/DCMAW-9394: save persistent id

alex-bradbury:
- Update secure store version and small refactor
- SecureStoreModule injects two versions of secure store
- SaveToOpenSecureStore created to access 'new' secure store
- GetPersistentId pulls persistent id from id token
- some refactoring of get token from token secure store
- logic in HomeScreenViewModel to extract and save persistent id

### DIFF
--- a/app/src/main/java/uk/gov/onelogin/tokens/usecases/GetPersistentId.kt
+++ b/app/src/main/java/uk/gov/onelogin/tokens/usecases/GetPersistentId.kt
@@ -18,7 +18,6 @@ class GetPersistentIdImpl @Inject constructor(
 ) : GetPersistentId {
     override fun invoke(): String? {
         val idToken: String = tokenRepository.getTokenResponse()?.idToken ?: return null
-        val id = idToken.extractPersistentIdFromIdToken()
-        return id
+        return idToken.extractPersistentIdFromIdToken()
     }
 }
--- a/app/src/main/java/uk/gov/onelogin/tokens/usecases/GetPersistentId.kt
+++ b/app/src/main/java/uk/gov/onelogin/tokens/usecases/GetPersistentId.kt
@@ -1,0 +1,24 @@
+package uk.gov.onelogin.tokens.usecases
+
+import javax.inject.Inject
+import uk.gov.onelogin.login.usecase.extractPersistentIdFromIdToken
+import uk.gov.onelogin.repositiories.TokenRepository
+
+fun interface GetPersistentId {
+    /**
+     * Use case to get the persistent id from the id token
+     *
+     * @return persistent id as a string or null if it fails to retrieve it
+     */
+    operator fun invoke(): String?
+}
+
+class GetPersistentIdImpl @Inject constructor(
+    val tokenRepository: TokenRepository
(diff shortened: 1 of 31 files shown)


app/src/main/java/uk/gov/onelogin/tokens/usecases/GetPersistentId.kt
```diff
@@ -0,0 +1,24 @@
...
+     * @return persistent id as a string or null if it fails to retrieve it
+     */
+    operator fun invoke(): String?
+}
+
+class GetPersistentIdImpl @Inject constructor(
+    val tokenRepository: TokenRepository
+) : GetPersistentId {
+    override fun invoke(): String? {
+        val idToken: String = tokenRepository.getTokenResponse()?.idToken ?: return null
```
flavius-mester, line 20: We might need to know if it fails to retrieve/extract the `persistentSessionID`?

app/src/main/java/uk/gov/onelogin/tokens/usecases/RemoveAllSecureStoreData.kt
```diff
@@ -1,34 +1,29 @@
...
         try {
-            secureStore.delete(
-                context = context,
+            tokenSecureStore.delete(
                 key = Keys.ACCESS_TOKEN_KEY
             )
-            secureStore.delete(
-                context = context,
+            tokenSecureStore.delete(
                 key = Keys.ID_TOKEN_KEY
```
flavius-mester, line 27: Do you need to remove the `PersistentSessionID` here?

alex-bradbury: It's not a part of this ticket, but eventually yes

app/src/main/java/uk/gov/onelogin/ui/home/HomeScreenViewModel.kt
```diff
@@ -10,31 +9,37 @@ import uk.gov.android.authentication.TokenResponse
...
                 saveTokenExpiry(tokenResponse.accessTokenExpirationTime)
                 saveToSecureStore(
-                    context = context,
                     key = Keys.ACCESS_TOKEN_KEY,
                     value = tokenResponse.accessToken
                 )
                 tokenResponse.idToken?.let {
+                    saveToOpenSecureStore(
+                        key = Keys.PERSISTENT_ID_KEY,
+                        value = getPersistentId() ?: ""
```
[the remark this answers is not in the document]
alex-bradbury: yeah fair point, happy to skip the save if empty

app/src/main/java/uk/gov/onelogin/tokens/usecases/GetPersistentId.kt
alex-bradbury: Well it will just be null if it fails to retrieve, this is the pattern for retrieving the email for example. It also allows for missing id token _or_ missing persistent id to fail in the same way

flavius-mester (approved):
👍

PR #185, opened 2024-08-21
feat: DCMAW-9390 send persistent id 

alex-bradbury:
Ticket: [https://govukverify.atlassian.net/browse/DCMAW-9390]

This update will send the previously saved Persistent ID in the `/authorize` call made via the authentication module.

- This include some refactoring due to updates to the authentication module. These are seen mostly in `MainActivityViewModel` and `WelcomeScreenViewModel`; moving from the old 'register for activity result` to creating an 'activity for result launcher'
- Created use case for accessing items in the 'open' secure store instance. 
- Created use case to get the persistent id from either the id token in working memory or failing that, the 'open' secure store
- Persistent Id is then send in as a parameter in the `LoginSessionConfiguration`


gradle/libs.versions.toml
```diff
@@ -69,7 +71,7 @@ pages = { group = "uk.gov.android", name = "pages", version.ref = "govuk-ui" }
 slf4j-api = { group = "org.slf4j", name = "slf4j-api", version = "2.0.16" }
 theme = { group = "uk.gov.android", name = "theme", version.ref = "govuk-ui" }
 uiautomator = { group = "androidx.test.uiautomator", name = "uiautomator", version = "2.3.0" }
-authentication = "uk.gov.android:authentication:0.5.0"
+authentication = "uk.gov.android.authentication:app:0.6.2"
```
ThomasIent, line 74: Could this be put into the same format as the other libraries i.e.
`authentication = { group = "uk.gov.android.authentication", name = "app", version.ref = "<version_ref>" }`?

app/src/main/java/uk/gov/onelogin/tokens/usecases/GetFromOpenSecureStore.kt
```diff
@@ -0,0 +1,41 @@
+package uk.gov.onelogin.tokens.usecases
+
+import android.util.Log
+import javax.inject.Inject
+import javax.inject.Named
+import uk.gov.android.securestore.RetrievalEvent
+import uk.gov.android.securestore.SecureStore
+
+fun interface GetFromOpenSecureStore {
```
ThomasIent, line 9: Turning Types and Objects into functions by overriding the invoke operator is pointless as functions are already treated as first class objects and can be passed around freely. It also causes the heinous naming convention where suddenly you have Objects, Types, and properties that should be called nouns named as verbs, which lowers the reading comprehension. 

Coding should be as simple as humanly possible, and "clever" or extremely language idiomatic practices should be avoided and only used as a last resort. I do understand that this is done elsewhere, so I don't expect you to change it, but I will be advocating these instances be changed on mass, if not individually.

Always willing to hear other arguments though.

app/src/test/java/uk/gov/onelogin/ui/welcome/WelcomeScreenViewModelTest.kt
```diff
@@ -0,0 +1,243 @@
...
+import uk.gov.onelogin.login.biooptin.BiometricPreference
+import uk.gov.onelogin.login.biooptin.BiometricPreferenceHandler
+import uk.gov.onelogin.login.ui.welcome.WelcomeScreenViewModel
+import uk.gov.onelogin.login.usecase.VerifyIdToken
+import uk.gov.onelogin.mainnav.nav.MainNavRoutes
+import uk.gov.onelogin.repositiories.TokenRepository
+import uk.gov.onelogin.tokens.usecases.AutoInitialiseSecureStore
+import uk.gov.onelogin.tokens.usecases.GetPersistentId
+
+@OptIn(ExperimentalCoroutinesApi::class)
```
ThomasIent, line 39: ~I've seen this a few times, should we not move this to the build.gradle.kts~
I was going to put something about potentially using freeCompilerArgs, but apparently they're going to deprecate it.

app/src/test/java/uk/gov/onelogin/ui/welcome/WelcomeScreenViewModelTest.kt
```diff
@@ -0,0 +1,243 @@
...
+        mockAutoInitialiseSecureStore,
+        mockVerifyIdToken,
+        mockFeatureFlags,
+        mockGetPersistentId,
+        mockOnlineChecker
+    )
+
+    @BeforeEach
+    fun setup() {
+        viewModel.next.observeForever(observer)
```
ThomasIent, line 79: No `@AfterEach fun tearDown()` to remover the observer? Considering the ViewModel is a value this would mean the observer would leak after each test, no?

app/src/test/java/uk/gov/onelogin/MainActivityViewModelTest.kt
```diff
@@ -57,13 +41,9 @@ class MainActivityViewModelTest {
...
-        mockLoginSession,
-        mockCredChecker,
         mockBioPrefHandler,
         mockTokenRepository,
-        mockAutoInitialiseSecureStore,
-        mockVerifyIdToken
+        mockAutoInitialiseSecureStore
     )
 
     @BeforeEach
```
ThomasIent, line 49: A LiveData `observeForever()` without the corresponding `removeObserver()`. Pairs of functions should have both available

app/src/test/java/uk/gov/onelogin/MainActivityViewModelTest.kt
```diff
@@ -57,13 +41,9 @@ class MainActivityViewModelTest {
     )
 
     private val viewModel = MainActivityViewModel(
```
ThomasIent, line 43: The system under test (MainActivityViewModel) should be recreated each fresh for each test, so that any possible state is not accidentally retained between tests. This is why it is still acceptable to use `lateinit var` within test classes, when it has fallen out of favour for production code.

ThomasIent (approved):
This is a good size piece of work that you have completed quickly that has a great level of testing, well done. There are a few improvements that need to be looked at in the long run, especially around testing, but we will need to agree on that as a team before acting on it.
I don't expect everything (or anything) I have raised to be change or "fixed" before this is merged, but I do expect it to be considered.

app/src/test/java/uk/gov/onelogin/MainActivityViewModelTest.kt
alex-bradbury: done

app/src/test/java/uk/gov/onelogin/MainActivityViewModelTest.kt
alex-bradbury: fixed

app/src/test/java/uk/gov/onelogin/ui/welcome/WelcomeScreenViewModelTest.kt
alex-bradbury: fixed

gradle/libs.versions.toml
alex-bradbury: agreed

app/src/main/java/uk/gov/onelogin/tokens/usecases/GetFromOpenSecureStore.kt
alex-bradbury: We can discuss this as it's quite a large feature of our codebase

PR #188, opened 2024-08-27
fix: update rexml version

alex-bradbury:


flavius-mester (approved):
subject to pipeline fix

PR #203, opened 2024-09-19
feat: Implement DeleteWalletData

BiancaMihaila:
**Ticket Link:** 
[DCMAW-9185](https://govukverify.atlassian.net/browse/DCMAW-9185)

**Description Of Changes:**
 - update java version used to v17
- update walletSdk version - update other dependencies to newer versions to be compatible with the wallet sdk
- update all use cases dependent on deleteWalletData
- update Wallet module: new wallet impl and delete data use case
- amend/ write tests accordingly

**Evidence:**
AC3 cannot de demo-ed - there is a unit test for it.

AC1:
Scenario 1: Credentials deleted successfully when Wallet API is called
GIVEN I am a user on the strategic app
AND I’m on the Sign out confirmation page
WHEN I tap the 'Sign out and delete app data' CTA
THEN Wallet credential deletion API is called
AND my Wallet credentials are deleted

https://github.com/user-attachments/assets/5816d597-be29-48f7-91b3-99aa3ac5dd3f

AC2:
Scenario 2 - User treated as ‘First time user’ when they return to the app (goes to Sign in page)
GIVEN I’m a logged in user on the Strategic app
AND I have credentials stored in my Wallet
AND I sign out using the functionality on the Profile tab
AND my sign out is successful
AND I leave the app
WHEN I return to the app
THEN I’m directed down the ‘First time user’ flow
AND I cannot see any Wallet credentials upon successful login

https://github.com/user-attachments/assets/5c816e88-4a37-4afe-8657-b244d79c13fa

**Definition Of Done Checklist:**

- [x] Unit tests written to at least 80% code coverage (unless explicit approval given from TL that this is not necessary)
- [x] Relevant integration and end-to-end tests are written
- [x] Tests cover each acceptance criteria on the user story
- [ ] Demo completed by running code locally, demoing to a code reviewer & designer
- [x] Sonar coverage gates met 
- [ ] If feature flagged, Feature is enabled in staging and manually tested to ensure integration

Resolves: DCMAW-9185

[DCMAW-9185]: https://govukverify.atlassian.net/browse/DCMAW-9185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/androidTest/java/uk/gov/onelogin/signOut/ui/SignOutScreenKtTest.kt
+++ b/app/src/androidTest/java/uk/gov/onelogin/signOut/ui/SignOutScreenKtTest.kt
@@ -7,26 +7,39 @@ import androidx.compose.ui.test.performClick
 import dagger.hilt.android.testing.BindValue
 import dagger.hilt.android.testing.HiltAndroidTest
 import dagger.hilt.android.testing.UninstallModules
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.test.runTest
 import org.junit.Before
 import org.junit.Test
 import org.mockito.Mockito.mock
+import org.mockito.kotlin.any
 import org.mockito.kotlin.never
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import uk.gov.android.onelogin.R
+import uk.gov.android.wallet.sdk.WalletSdk
 import uk.gov.onelogin.TestCase
 import uk.gov.onelogin.signOut.SignOutModule
 import uk.gov.onelogin.signOut.domain.SignOutError
 import uk.gov.onelogin.signOut.domain.SignOutUseCase
+import uk.gov.onelogin.wallet.DeleteWalletDataUseCase
+import uk.gov.onelogin.wallet.WalletModule
 
 @HiltAndroidTest
 @UninstallModules(
-    SignOutModule::class
+    SignOutModule::class,
+    WalletModule::class
 )
 class SignOutScreenKtTest : TestCase() {
 
     @BindValue
     val signOutUseCase: SignOutUseCase = mock()
+
+    @BindValue
+    val walletSdk: WalletSdk = mock()
+
+    @BindValue
+    val deleteWalletDataUseCase: DeleteWalletDataUseCase = mock()
     private val goBack: () -> Unit = mock()
     private val signIn: () -> Unit = mock()
     private val goToSignOutError: () -> Unit = mock()
@@ -49,18 +62,18 @@ class SignOutScreenKtTest : TestCase() {
     }
 
     @Test
-    fun verifySignOutButtonSucceeds() {
+    fun verifySignOutButtonSucceeds() = runBlocking {
         composeTestRule.onNode(ctaButton).performClick()
-        verify(signOutUseCase).invoke()
+        verify(signOutUseCase).invoke(any())
         verify(signIn).invoke()
     }
 
-    @Test
-    fun verifySignOutButtonFails() {
-        whenever(signOutUseCase.invoke())
+    @Test(expected = Throwable::class)
+    fun verifySignOutButtonFails() = runTest {
+        whenever(signOutUseCase.invoke(any()))
             .thenThrow(SignOutError(Exception("something went wrong")))
         composeTestRule.onNode(ctaButton).performClick()
-        verify(signOutUseCase).invoke()
+        verify(signOutUseCase).invoke(any())
         verify(signIn, never()).invoke()
         verify(goToSignOutError).invoke()
     }
--- a/app/src/main/java/uk/gov/onelogin/mainnav/ui/MainNavScreen.kt
+++ b/app/src/main/java/uk/gov/onelogin/mainnav/ui/MainNavScreen.kt
@@ -96,7 +96,7 @@ fun MainNavScreen(
                 HomeScreen(openDeveloperPanel = openDeveloperPanel)
             }
             composable(BottomNavDestination.Wallet.key) {
-                walletScreenViewModel.walletSdk.WalletApp(deeplink = "", adminEnabled = true)
+                walletScreenViewModel.walletSdk.WalletApp(deeplink = "", adminEnabled = false)
             }
             composable(BottomNavDestination.Profile.key) {
                 ProfileScreen(openSignOutScreen = openSignOutScreen)
--- a/app/src/main/java/uk/gov/onelogin/signOut/domain/SignOutUseCase.kt
+++ b/app/src/main/java/uk/gov/onelogin/signOut/domain/SignOutUseCase.kt
@@ -24,7 +24,7 @@ class SignOutUseCaseImpl @Inject constructor(
             removeTokenExpiry()
             removeAllSecureStoreData()
             bioPrefHandler.clear()
-            deleteWalletData(activityFragment)
+            deleteWalletData.invoke(activityFragment)
         } catch (e: Throwable) {
             throw SignOutError(e)
         }
--- a/app/src/main/java/uk/gov/onelogin/signOut/domain/SignOutUseCase.kt
+++ b/app/src/main/java/uk/gov/onelogin/signOut/domain/SignOutUseCase.kt
@@ -8,7 +8,7 @@ import uk.gov.onelogin.tokens.usecases.RemoveTokenExpiry
 import uk.gov.onelogin.wallet.DeleteWalletDataUseCase
 
 fun interface SignOutUseCase {
-    suspend operator fun invoke(activityFragment: FragmentActivity)
+    suspend fun invoke(activityFragment: FragmentActivity)
 }
 
 @Suppress("TooGenericExceptionCaught")
--- a/app/src/main/java/uk/gov/onelogin/signOut/domain/SignOutUseCase.kt
+++ b/app/src/main/java/uk/gov/onelogin/signOut/domain/SignOutUseCase.kt
@@ -1,26 +1,30 @@
 package uk.gov.onelogin.signOut.domain
 
+import androidx.fragment.app.FragmentActivity
 import javax.inject.Inject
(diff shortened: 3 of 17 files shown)


app/src/main/java/uk/gov/onelogin/wallet/DeleteWalletDataUseCase.kt
```diff
@@ -0,0 +1,31 @@
+package uk.gov.onelogin.wallet
+
+import androidx.fragment.app.FragmentActivity
+import javax.inject.Inject
+import uk.gov.android.wallet.sdk.WalletSdk
+
+fun interface DeleteWalletDataUseCase {
+    suspend operator fun invoke(activityFragment: FragmentActivity)
```
ThomasIent, line 8: Please create a named function that represents what is going on (clearWalletData?) and don't override the invoke operator. I understand that in our current usecase the invoke operator is used, but going forward we will be moving towards clean coding, of which this violates a significant number of principles (none of which are hard rules, but are there to point the correct direction), including:

- Naming convention, Verbs for actions: as invoke is not required at the calling point, suddenly you have (or can have) a noun (class name) performing an action instead of a verb (function name)
- Naming convention, functions named for action to be performed: Invoke tells us what is happening with the class, but gives no indication to the calling site about the action that is actually being performed, as it is overwritten
- Throws exception in constructor: It is convention (in all OO languages) to forbid exceptions being thrown in constructors, which in Kotlin extends to init blocks and invoke calls, as these are called at object construction
- Idiomatic coding: Any programmer familiar with OO should be able to read the code we write (especially us in the future), overriding the invoke operator is extremely specific to Kotlin, and causes confusion as you suddenly have object construction used as a function call `val objectA = SomeObject()`, should be reserved for actual object creation and not performing 'hidden' actions when an action is not expected

There are others, but that is what I have off the top of my head, without having to go get a text book.
It is one of those things that is very 'clever', but we don't want clever, we want dumb / simple / easy

app/src/main/java/uk/gov/onelogin/wallet/DeleteWalletDataUseCase.kt
```diff
@@ -0,0 +1,31 @@
...
+    "RethrowCaughtException",
+    "TooGenericExceptionThrown"
+)
+class DeleteWalletDataUseCaseImpl @Inject constructor(
+    private val walletSdk: WalletSdk
+) : DeleteWalletDataUseCase {
+    override suspend operator fun invoke(activityFragment: FragmentActivity) {
+        try {
+            val deleteResult = walletSdk.deleteWalletData(activityFragment)
+            if (!deleteResult) throw Exception(DELETE_WALLET_DATA_ERROR)
```
ThomasIent, line 22: Please create a specific Exception to be thrown here

ThomasIent: Also, the big red warnings you've suppressed are there to help you. They're not there to get in your way, but to guide you to a cleaner / better solution that will help you in the long run. If you are suppressing warnings, then I would like a written dev comment detailing why, please 

P.S. expect the same from others, if you see us just suppress things, please call it out.

app/src/androidTest/java/uk/gov/onelogin/signOut/ui/SignOutScreenKtTest.kt
```diff
@@ -7,26 +7,39 @@ import androidx.compose.ui.test.performClick
...
+import uk.gov.onelogin.wallet.DeleteWalletDataUseCase
+import uk.gov.onelogin.wallet.WalletModule
 
 @HiltAndroidTest
 @UninstallModules(
-    SignOutModule::class
+    SignOutModule::class,
+    WalletModule::class
 )
 class SignOutScreenKtTest : TestCase() {
```
ThomasIent, line 33: _I know this isn't something you have done_
Test classes should be named <ClassName>Test, the Kt bit gets added by mistake via Android Studio. Do watch out for this as you make tests, and it is appreciated, but not essential, if you could rename them when you come across them while making changes (should be in a separate commit just for the renaming if you do please). 
Though leave it alone if not making any changes to that file

app/src/test/java/uk/gov/onelogin/signOut/domain/SignOutUseCaseTest.kt
```diff
@@ -35,8 +49,19 @@ class SignOutUseCaseTest {
...
         val exception: SignOutError = Assertions.assertThrows(SignOutError::class.java) {
-            sut.invoke()
+            runBlocking { sut.invoke(activityFragment) }
         }
         assertTrue(exception.error.message!!.contains("something went terribly bad"))
     }
+
+    @Test
+    fun `sign out delete wallet data error`() = runBlocking {
+        whenever(deleteWalletData.invoke(activityFragment)).then {
```
ThomasIent, line 59:
```
whenever(deleteWalletData.invoke(activityFragment)).thenThrow(Exception(DELETE_WALLET_DATA_ERROR))
```
Will get rid of the file warning, and thus allow you to remove the suppression

app/src/test/java/uk/gov/onelogin/wallet/DeleteWalletDataUseCaseTest.kt
```diff
@@ -0,0 +1,50 @@
...
+    fun `walled sdk init error`() = runBlocking {
+        whenever(walletSdk.deleteWalletData(any())).then {
+            throw WalletSdk.WalletSdkError.SdkNotInitialised
+        }
+        val result = Assertions.assertThrows(
+            WalletSdk.WalletSdkError.SdkNotInitialised::class.java
+        ) {
+            runBlocking { sut.invoke(activityFragment) }
+        }
+        assertTrue(result.instanceOf(WalletSdk.WalletSdkError.SdkNotInitialised::class))
```
ThomasIent, line 48: This assertTrue is a duplication of assertThrows, so you could have 
```
Assertions.assertThrows(
            WalletSdk.WalletSdkError.SdkNotInitialised::class.java
        ) {
            runBlocking { sut.invoke(activityFragment) }
        }
```
without any result set

app/src/test/java/uk/gov/onelogin/wallet/DeleteWalletDataUseCaseTest.kt
```diff
@@ -0,0 +1,50 @@
+package uk.gov.onelogin.wallet
+
+import androidx.fragment.app.FragmentActivity
+import io.ktor.util.reflect.instanceOf
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlinx.coroutines.runBlocking
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Test
```
ThomasIent, line 9: Note for the future: Please favour the Kotlin test imports. 
It is a bit subjective, but if we need to swap / downgrade the JUnit version (we actually might!) then we don't have to change any tests as Kotlin test is library agnostic

app/src/main/java/uk/gov/onelogin/mainnav/ui/MainNavScreen.kt
```diff
@@ -96,7 +96,7 @@ fun MainNavScreen(
                 HomeScreen(openDeveloperPanel = openDeveloperPanel)
             }
             composable(BottomNavDestination.Wallet.key) {
-                walletScreenViewModel.walletSdk.WalletApp(deeplink = "", adminEnabled = true)
+                walletScreenViewModel.walletSdk.WalletApp(deeplink = "", adminEnabled = false)
```
ThomasIent, line 99: _I know this isn't something you have done_
The naming convention for booleans is to prepend them with `is`, `has`, or `should`, so they read grammatically correct when used in conditionals or as predicates. The same is true for functions that return booleans, as well as boolean variables.

app/src/main/java/uk/gov/onelogin/signOut/domain/SignOutUseCase.kt
```diff
@@ -1,26 +1,30 @@
...
+import androidx.fragment.app.FragmentActivity
 import javax.inject.Inject
 import uk.gov.onelogin.login.biooptin.BiometricPreferenceHandler
 import uk.gov.onelogin.tokens.usecases.RemoveAllSecureStoreData
 import uk.gov.onelogin.tokens.usecases.RemoveTokenExpiry
+import uk.gov.onelogin.wallet.DeleteWalletDataUseCase
 
 fun interface SignOutUseCase {
-    operator fun invoke()
+    suspend operator fun invoke(activityFragment: FragmentActivity)
```
ThomasIent, line 11: Please refer to my large rant in DeleteWalletDataUsecase, as the same applies here. You could argue that it was already in use and therefore keeping it that way is ok, but I believe in refactoring as you go, and as changes are being made then the new code should better than the old code.

ThomasIent (changes requested):
Overall a really good piece of work. 
Do remember that commits should have separation of concerns just as code has separation of concerns, so I would like you to split the commits into the java 17 updates and then the work, please. 
I would also like functions in the two usecases I commented on to be made separate from the invoke operator, please.

app/src/main/java/uk/gov/onelogin/wallet/DeleteWalletDataUseCase.kt
ThomasIent: Pulling out discussion on "overriding the invoke" to a tech debt ticket, so it can be discussed in more detail.

ThomasIent (approved):

ThomasIent:
I am happy for this to be reviewed by QA.

co-eoin:
QA Reviewed & Approved ✅
Splash screen issue seen in AC2 has been reported as a [defect ](https://govukverify.atlassian.net/browse/DCMAW-10260) (Thank you Tom).

PR #204, opened 2024-09-20
Fix/refactor navigation

alex-bradbury-gds:
Creation of `Navigator`:
- This interface allows us to have an 'injectable' handle on navigation
- It can be 'initialised' via the setController method
- Then a number of 'standard' navigation methods; `navigate()`, `goBack()`, `hasBackStack()` and `openDeveloperPanel()` allow us to make these calls directly from ViewModels
- The `reset()` method can be use to 'unset' the NavController to avoid leakage.
- In the `MainActivity` the navigator is essentially tied to the lifecycle of the main composable (basically the activity's lifecycle)

Creation of `NavRoute` interface:
- Allows a type-safe way of passing navigation routes
- Can see the 'conversion' of many of the existing 'routes' to new sealed classes that implement the `NavRoute`

General comments:
- This has allowed the moving of navigation logic out of the 'Routes' files, which create the compose navigation sub-graphs, and into the ViewModels. Most of the work here is converting our existing setup into this approach
- this commit has only covered the conversion of the Login routes
- Convert ErrorGraph, MainNavGraph and SignOutGraph to create routes extending the NavRoute interface
- Fix developer panel navigation
- Remove all compose lambdas from screens that have a viewmodel which can instead handle the navigation


app/src/main/java/uk/gov/onelogin/navigation/NavigatorImpl.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import android.util.Log
+import androidx.navigation.NavHostController
+import javax.inject.Inject
+import uk.gov.onelogin.developer.DeveloperRoutes.navigateToDeveloperPanel
+
+class NavigatorImpl @Inject constructor() : Navigator {
+    private var navController: NavHostController? = null
+        get() {
+            if (field == null) Log.w(this::class.simpleName, "Navigator not initialised")
```
BiancaMihaila, line 11: What happens if the navController is not initialised here? Would there be any situation where this would be possible?

alex-bradbury-gds: There shouldn't be, it should be linked to the lifecycle of the Activity. If `navigate()` is called and it is not set then it will at least make a warning log, it won't crash the app, it simply just won't navigate.

app/src/main/java/uk/gov/onelogin/login/ui/biooptin/BiometricsOptInScreen.kt
```diff
@@ -30,14 +30,11 @@ import uk.gov.android.ui.theme.GdsTheme
 import uk.gov.android.ui.theme.hintTextGrey
 import uk.gov.android.ui.theme.m3.Typography
 import uk.gov.android.ui.theme.smallPadding
-import uk.gov.onelogin.login.biooptin.BiometricPreference
 
 @Suppress("LongMethod")
 @Composable
 fun BiometricsOptInScreen(
```
ThomasIent, line 36: It's definitely a nice refactor when responsibility is moved and the parameters start disappearing

app/src/main/java/uk/gov/onelogin/navigation/NavigatorImpl.kt
ThomasIent: I would argue that it would be better to have a crash, as I think of it as:
- A problem a developer needs to be aware of -> crash
- A problem a user needs to be aware of -> Recoverable path with Error Screen 
but I don't actually have a problem with how you have done this. I think it is fine

app/src/main/java/uk/gov/onelogin/login/LoginGraphObject.kt
```diff
@@ -0,0 +1,99 @@
...
+object LoginGraphObject {
+    @Suppress("LongMethod")
+    fun NavGraphBuilder.loginGraph(
+        navController: NavHostController
+    ) {
+        navigation(
+            route = LoginRoutes.Root.getRoute(),
+            startDestination = LoginRoutes.Start.getRoute()
+        ) {
+            composable(
```
ThomasIent, line 28: I wonder if it is best to split each navigation block composable into its own function to reduce the length of this function and make it easier to read. Something to look into at a later date maybe.

ThomasIent (approved):
Really nice bit of work.

app/src/main/java/uk/gov/onelogin/login/LoginGraphObject.kt
alex-bradbury-gds: Yeah agreed, definitely could be a nice idea. Will probably not include that here though, bit overloaded already in this PR

PR #226, opened 2024-10-03
feat: Add ReAuthErrorScreen and ViewModel 

BiancaMihaila:


app/src/main/java/uk/gov/onelogin/ui/error/reauth/ReAuthErrorScreen.kt
```diff
@@ -0,0 +1,180 @@
+package uk.gov.onelogin.ui.error.reauth
```
BiancaMihaila, line 1: Wasn't sure if this specific layout for a component was going to be used in other places - but I can move the code in mobile-ui repo and make it a bit more re-usable if it's easier to do it now.

PR #228, opened 2024-10-07
feat: DCMAW-9193 checks for persistent id

alex-bradbury-gds:
## Description

- Signed out screen now checks for persistent id and then deletes data and routes accordingly if no persistent id is found

## Evidence

AC 1 and 3

https://github.com/user-attachments/assets/a9bf430e-2446-4048-b79f-15c8f0dffd9b

AC 2 : Some information here, initial login was with abradbury@.... email. The following login is using a different email (as evidenced), that email does not actually have an account but the expected behaviour if I did have a PersisitentSessionID would be to see a prompt to enter 'the same email as before', which we do not see here

https://github.com/user-attachments/assets/9a1fd1b8-16d0-4c54-8da7-78ac957f6a53

AC Error 1:

https://github.com/user-attachments/assets/424da75d-75df-4611-8343-f667a8f192ce

AC Regression 1:

https://github.com/user-attachments/assets/b59532a6-e964-4822-9a1d-2975e26ee1cf

[DCMAW-9193]: https://govukverify.atlassian.net/browse/DCMAW-9193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BiancaMihaila (approved):

co-eoin:
QA Reviewed & Approved!

PR #307, opened 2025-01-02
refactor: WelcomeScreenViewModel

alex-bradbury-gds:
The main idea is to break up the complexity of the `WelcomeScreenViewModel`. This refactor hasn't reduced the constructor parameters tooo much unfortunately however the general complexity of the view model has been reduced.

- Remove `LoginSession` calls from `WelcomeScreenViewiodel`
- HandLeRemoteLogin.kt handles logic to do with starting the signing process via the *LoginSession •present()• and getting the client attestation
- HandleLoginRedirect.kt handles the response back from the login redirect, sets up the necessary bits for app integrity (attestation and PoP) to attach to the /token call to complete login
- Uri parser created for ease of testing

### DIFF
--- a/app/src/main/java/uk/gov/onelogin/login/ui/welcome/WelcomeScreenViewModel.kt
+++ b/app/src/main/java/uk/gov/onelogin/login/ui/welcome/WelcomeScreenViewModel.kt
@@ -2,7 +2,6 @@ package uk.gov.onelogin.login.ui.welcome
 
 import android.content.Context
 import android.content.Intent
-import android.net.Uri
 import android.util.Log
 import androidx.activity.result.ActivityResultLauncher
 import androidx.lifecycle.ViewModel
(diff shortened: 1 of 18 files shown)


app/src/main/java/uk/gov/onelogin/login/ui/welcome/WelcomeScreenViewModel.kt
```diff
@@ -13,190 +12,75 @@ import javax.inject.Inject
...
+                onFailure = {
+                    Log.e(tag, it?.message, it)
                     navigator.navigate(LoginRoutes.SignInError, true)
                 }
-            }
-        } else {
-            callback("")
+            )
         }
     }
```
BiancaMihaila, line 86: You mentioned earlier but we could move the `handleTokens()` into a use case similar to handle login (including the handle tokens fun and check local auth route)?

BiancaMihaila (approved):

app/src/main/java/uk/gov/onelogin/login/ui/welcome/WelcomeScreenViewModel.kt
alex-bradbury-gds: Yeah definitely, that would be the next think to do i think. Will probably raise a fresh PR for that

ThomasIent (approved):
Looks good

PR #456, opened 2025-05-12
DCMAW-11704: Add branding to the home tab title banner

chrisdondigital:
- Add OneLogin image to home screen top app bar
- Add correct day and night colours to top app bar
- Add content description to OneLogin image
- Enable edge to edge padding to all screens


alex-bradbury-gds (commented):
Is it not possible to wrap the app at a top level with `EdgeToEdgePage` rather than on each composable?

chrisdondigital:
> Is it not possible to wrap the app at a top level with `EdgeToEdgePage` rather than on each composable?

I don't think it is because in some screens we don't want the padding applied (like wallet and settings screen). Also, applying in one place would involve wrapping a Scaffold around the NavHost which doesn't feel right.

alex-bradbury-gds (approved):

PR #462, opened 2025-05-15
feat: update to latest CRI and Secure Store versions

alex-bradbury-gds:
- Update logout behavior to be linked to application lifecycle
- Bump CRI version
- Bump Secure Store version
- Update CRI feature flag to be `Enabled` by default


app/build.gradle.kts
```diff
@@ -226,11 +225,13 @@ dependencies {
         libs.mockito.kotlin,
         libs.junit.jupiter,
         libs.junit.jupiter.params,
+        libs.junit.vintage.engine,
```
BiancaMihaila, line 228: shouldn't this be only applied as a runtimeImplementation?

BiancaMihaila (approved):

PR #476, opened 2025-05-20
DCMAW-12636: Focus sequence disorientated 

chrisdondigital:
- Use traversalIndex to correctly order elements in settings screen when using a screen reader


core/src/main/java/uk/gov/onelogin/core/ui/components/EmailSection.kt
```diff
@@ -66,3 +69,4 @@ internal fun EmailSectionPreview() {
 
 const val DIVIDER_TEST_TAG = "divider_test_tag"
 internal const val IMAGE_TEST_TAG = "image_test_tag"
+private const val EMAIL_TRAVERSAL_ORDER = -25f
```
BiancaMihaila, line 72: just out of curiosity - I'm assuming you're using this to override the transversal order for TalkBack? (just asking because I looked at this at some point and I tried it with different numbers but wasn't behaving as I expected)

BiancaMihaila (approved):

core/src/main/java/uk/gov/onelogin/core/ui/components/EmailSection.kt
chrisdondigital: That's right. The default value is zero but the lowest value index number gets priority. I had to play around with the numbers to get it to work because it does appear to defy logic!

PR #530, opened 2025-06-12
fix: add sdk app name

emanuel-gds:
# DCMAW-12794

- Add the `appName` required in ID Check SDK

https://github.com/govuk-one-login/mobile-id-check-android-sdk/blob/66304bec3709662400251e684b1f008626ebbd15/modules/idcheck/network-api/src/main/res/values/strings.xml

The plan is to name space the id check sdk strings so this will change to `idsdk_app_name` in the future


emanuel-gds:
Discussed with Alex and this should be defined in CRI orchestrator

PR #550, opened 2025-06-24
release/1.0.6 - Amend versioning on release [TEST - DO NOT MERGE/ APPROVE]

BiancaMihaila:


.github/workflows/on_push-develop.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         uses: ./config/actions/version-name
         with:
           version-code: ${{ steps.version-code.outputs.version-code }}
-          version-name: ${{ steps.latest-tag.outputs.current-tag }}
+          version-name: ${{ steps.get-latest-tag.outputs.current-tag }}
```
BiancaMihaila, line 62: I will revert this I don't think ti makes any difference - I noticed that the new versions on develop seem to now be publishing with the correct tag - I looked at the what Chris merged yesterday and it seems it published with 1.1.2 as part of it, but I'll run a couple more on https://github.com/govuk-one-login/mobile-android-one-login-app/pull/551

.github/workflows/on_push-main.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Get next version from branch
+        id: next-version
+        uses: ./config/actions/get-next-version-from-branch
+
+#      - name: Ensure version is correct
+#        id: version-number
+#        uses: ./config/actions/ensure-version-is-correct
+#        with:
+#          current-version: ${{ steps.latest-tag.outputs.current-tag }}
+#          next-version: ${{ steps.next-version.outputs.next-version }}
```
BiancaMihaila, line 60: This is commented only because it will throw an error since I'm using a version lower than the latest 1.1.2 - will uncomment if this works

BiancaMihaila:
This was for testing purposes only - will be closing it now that it has been completed.

PR #578, opened 2025-07-10
ci: Amend variable name in action to handle it correctly

BiancaMihaila:


BiancaMihaila:
This was added in the 1.3.0 release

PR #588, opened 2025-07-14
Unrecoverable sign in error screen

chrisdondigital:
[DCMAW-11137](https://govukverify.atlassian.net/browse/DCMAW-11137): Create an additional copy of the Sign in error page to cater for unrecoverable error scenarios
[DCMAW-13358](https://govukverify.atlassian.net/browse/DCMAW-13358): Add GA4 tracking to the sign in error page - unrecoverable

- Create unrecoverable error screen
- Add new login route
- Add logic for displaying correct screen
- Add new GA4 screen event

### Evidence
OAuth error generates a 500 http response code type server_error from the /authorise endpoint which results in the recoverable error screen being displayed
[DCMAW-11137-recov-evidence.webm](https://github.com/user-attachments/assets/4e21a160-28e8-4ea4-a11d-c04d37183e96)

Code changed to simulate a 400 error from the /token endpoint which results in the unrecoverable error screen being displayed
https://github.com/user-attachments/assets/6f816970-f388-4636-ad69-89c20327619a

[DCMAW-11137]: https://govukverify.atlassian.net/browse/DCMAW-11137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DCMAW-13358]: https://govukverify.atlassian.net/browse/DCMAW-13358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/test/java/uk/gov/onelogin/core/navigation/domain/NavControllerExtTest.kt
```diff
@@ -39,7 +39,7 @@ class NavControllerExtTest : FragmentActivityTestCase() {
         composeTestRule.runOnUiThread {
             // Fill up backstack
             navController.setCurrentDestination(LoginRoutes.Loading.getRoute())
-            navController.setCurrentDestination(LoginRoutes.SignInError.getRoute())
+            navController.setCurrentDestination(LoginRoutes.SignInRecoverableError.getRoute())
             navController.setCurrentDestination(LoginRoutes.AnalyticsOptIn.getRoute())
             navController.closeApp()
         }
```
BiancaMihaila, line 45: Could we add one in here for the UnrecoverableError as well please?

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/loginredirect/HandleLoginRedirectImpl.kt
```diff
@@ -49,45 +48,28 @@ class HandleLoginRedirectImpl @Inject constructor(
...
-                        e.message ?: NO_MESSAGE,
-                        e
-                    )
-                    onFailure(e)
-                }
+        loginSession.finalise(
+            intent = intent,
+            appIntegrity = AppIntegrityParameters(attestation, jwt),
+            { tokens ->
+                onSuccess(tokens)
```
BiancaMihaila, line 62: Sorry, just looked at the code in auth, and I think there's an `IllegalArgumentException` that can be thrown here (from auth, when attempting to validate the tokens) - if we don't wanna amend that now in auth repo, we might need to wrap the keep the finalise within a `try catch`

chrisdondigital: What throws the IllegalArgumentException? My understanding is that it's only AuthorizationResponse.fromIntent(intent) in loginSession.finalise which is now within a try catch. Therefore all errors will now be passed in through the onFailure callback.

core/src/test/java/uk/gov/onelogin/core/navigation/domain/NavControllerExtTest.kt
chrisdondigital: Will do

PR #606, opened 2025-07-23
release/1.4.0

BiancaMihaila:
- fix login issue: when clientAttestation is expired, we are attempting to retrieve a new one
- fix login issue: proof of possession seems to expire by the time it reaches back-end which blocks login, a change tp use java.time.Instant.now() instead of System.now() has been added and some error and info logs have been added to assist with monitoring this issue
- added feature that handles both recoverable and unrecoverable errors on Sign-in which includes changes to authentication that should now catch all/ most of the errors thrown and reduce the number of crashes we see in Prod

Resolves: DCMAW-11137, DCMAW-13358, DCMAW-14514, DCMAW-14540


BiancaMihaila:
Close because it's merging into wrong branch

PR #613, opened 2025-08-01
Release/1.5.0

BiancaMihaila:
- merge branch 'main' into feat/DCMAW-14672-reauth-when-adding-a-cred-when-token-expired
- bump wallet version to include the fix for the in app QR scanner when the access token is
expire (this includes running the navigation on the main thread)
- bump the wallet version to `0.24.0`
- bump cri-orch to `0.59.2` to include the fix for crash in prod
- - set the expiry token to 30 seconds form the button press
- this allows for testers to navigate to the wallet and not be prompted
for re-auth from CRI Orch


core/src/main/res/values/strings.xml
```diff
@@ -138,6 +142,7 @@
     <string name="app_developer_features_title" translatable="false">Select Features</string>
 
     <string name="app_developer_reset_access_token_button" translatable="false">Set Access Token Expiry to Now</string>
+    <string name="app_developer_access_token_button_expire_45_sec" translatable="false">Set Access Token to Expire in 30 Seconds</string>
```
alex-bradbury-gds, line 145: id tag says 45 seconds, but the text says 30 seconds...

PR #616, opened 2025-08-06
fix: Fix deeplink not working properly when app is in memory

BiancaMihaila:
# [fix: Fix deeplink not working properly when app is in memory ](https://govukverify.atlassian.net/browse/DCMAW-14763)

This issue would only happen when the app is open (int he background) and a credential has already been added and removed from he Wallet (it seems the deep link was persisted)

**Changes:**
- add functionality to clear the deep link and credential off saved in memory after it's been passed to the `WalletSdk`
- add a check to `finish()` the intent when `onNewIntent` is called - this allows to reset the previously existing intent - this is done only when the user comes in via deep link to add a Wallet credential
- separate the `WalletScreenViewModel` and the `MainNavViewModel` to allow for specific behaviour that allows the user to add a credential when the app is in memory

## Evidence

https://github.com/user-attachments/assets/0188ed6a-eb1a-4f4b-8a62-4a5c892577c1

Resolves: DCMAW-14763

### DIFF
--- a/app/src/main/java/uk/gov/onelogin/MainActivity.kt
+++ b/app/src/main/java/uk/gov/onelogin/MainActivity.kt
@@ -44,6 +44,6 @@ class MainActivity : AppCompatActivity() {
     override fun onNewIntent(intent: Intent) {
         super.onNewIntent(intent)
         setIntent(intent)
-        viewModel.handleIntent(intent)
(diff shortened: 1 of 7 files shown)


co-eoin:
QA reviewed & Approved

PR #641, opened 2025-09-08
fix: Save tokens in secure store when toggling biometrics on

BiancaMihaila:
- save tokens only when the state is from Disabled to Biometrics Enabled
- if from passcode or disabling biometrics the behaviour stays the same

**Evidence:**

https://github.com/user-attachments/assets/2ecd51dd-238e-41d2-aa6b-18255e57934f

Resolves: DCMAW-14594

### DIFF
--- a/features/src/main/java/uk/gov/onelogin/features/settings/ui/biometricstoggle/BiometricsToggleScreenViewModel.kt
+++ b/features/src/main/java/uk/gov/onelogin/features/settings/ui/biometricstoggle/BiometricsToggleScreenViewModel.kt
@@ -1,22 +1,26 @@
 package uk.gov.onelogin.features.settings.ui.biometricstoggle
 
 import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
 import kotlinx.coroutines.flow.MutableStateFlow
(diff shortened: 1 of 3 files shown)


features/src/main/java/uk/gov/onelogin/features/settings/ui/biometricstoggle/BiometricsToggleScreenViewModel.kt
```diff
@@ -34,6 +38,16 @@ class BiometricsToggleScreenViewModel @Inject constructor(
     }
 
     fun toggleBiometrics() {
+        // only if the initial local auth preference is disabled
+        if (localAuthManager.localAuthPreference is LocalAuthPreference.Disabled) {
+            // then we save the tokens in memory
+            // because the toggle has already been pressed AND this should never show unless the
+            // biometrics are available as we do a check when landing on teh screen (see checkBiometricsAvailable() call point)
+            viewModelScope.launch {
+                saveTokens()
```
alex-bradbury-gds, line 47: Has the secure store been properly initialised with the 'new' LocalAuthPref by this point?

BiancaMihaila: I've changed it now - thank you for helping with this 🙏

alex-bradbury-gds (approved):

PR #659, opened 2025-09-26
build: Update CRI Orchestrator to 0.61.7

jonnyandrew:


BiancaMihaila:
Can I check this version was QAd (regression) in ID-Check? (the reason I'm asking: https://gds.slack.com/archives/C0910G7BN58/p1758636384126949)
an we make sure that changing to API 36 doesn't affect Wallet and if so, happy to approve this - but if not, might require some more testing/ work depending on that. Can I check, which version of CRI Orch is the one requiring to upgrade to API 36 (it seems 0.61.5 works fine without it, but I might have missed something)?

PR #682, opened 2025-10-29
ci: Split workflow actions to run in parallel

BiancaMihaila:


BiancaMihaila:
Close this as multiple changes took place since this was started - will be replaced with a different PR/ branch.

PR #781, opened 2026-03-30
refactor: Split unit and component tests

BiancaMihaila:
- refactor tests to be split in unit and components
- rename OptIn package into AnalyticsOptIn to be more accurate/ specific
- update screenshots (been moved so will have a different name but no UI differences)

Resolves: DCMAW-19348


.github/workflows/on_pull_request.yml
```diff
@@ -22,7 +22,7 @@ concurrency:
 jobs:
   on-pull-request-job:
     name: Verify code base when pull request is published/updated
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
alex-bradbury-gds, line 25: is there a reason for pinning specific version?

config/actions/gradle-checks/action.yml
```diff
@@ -0,0 +1,78 @@
...
+    required: true
+  app-check-token:
+    description: 'App check token'
+    required: false
+    default: ''
+
+runs:
+  using: "composite"
+  steps:
+    - name: Run Tests
```
alex-bradbury-gds, line 28:
```suggestion
    - name: Run Unit Tests
```

alex-bradbury-gds (approved):

.github/workflows/on_pull_request.yml
BiancaMihaila: Trying to see if it fixes the issue - apparently 24.04 might be dealing differently with dependency resolution due to the Java version used, but I doubt will change much

config/actions/gradle-checks/action.yml
BiancaMihaila: Can I commit and amend this in teh next PR? Scared this will fail if I push any changes 😶‍🌫️

PR #815, opened 2026-04-16
build: Force versions for transitive dependencies to resolve security vulnerabilities

BiancaMihaila:


BiancaMihaila:
This has been tested in Sharing codebase and it does not solve the issue of vulnerabilities form 3rd party libs so will close

PR #819, opened 2026-04-20
ci: Split workflow in parallel jobs

BiancaMihaila:


.github/workflows/on_push_release.yml
```diff
@@ -73,7 +73,7 @@ jobs:
           CURRENT_VERSION: ${{ steps.latest-tag.outputs.current-tag }}
           NEXT_VERSION: ${{ steps.version-number.outputs.version }}
 
-  build-and-publish-release:
+  build-and-publish-release-build:
```
jonnyandrew: Could a job matrix be useful to reduce duplication between our build and publish jobs (both here and in the the other publishing workflows)?

It should allow us to run the same job for different environments:

https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations

.github/workflows/on_pull_request.yml
```diff
@@ -97,84 +133,274 @@ jobs:
...
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          lfs: 'true'
+          submodules: 'true'
+          fetch-depth: 0
+
+      - name: Detect Arch
+        id: detect-arch
+        uses: ./config/actions/detect-arch
+
```
jonnyandrew: We use `actions/detect-arch` across lots of the jobs here, but I can't see the `arch-in-use` output being used anywhere.

Is it needed, or could we drop it?

.github/workflows/run_checks.yml
```diff
@@ -97,84 +133,274 @@ jobs:
...
+        uses: ./config/actions/detect-arch
+
+      - name: Setup GitHub Runner workflow
+        uses: ./mobile-android-pipelines/actions/setup-runner
+
+      - name: Clean runner
+        uses: ./mobile-android-pipelines/actions/clean-runner-fast
+
+      - name: Restore coverage reports to build directories
+        uses: ./mobile-android-pipelines/actions/restore-coverage-reports
```
jonnyandrew: At first I wasn't sure what exactly 'restore' meant in this context.

Would it make sense to rename this action to `mobile-android-pipelines/actions/download-coverage-reports`? 

This also makes the naming symetrical with `mobile-android-pipelines/actions/upload-coverage-reports`.

.github/workflows/on_push_develop.yml
```diff
@@ -26,27 +26,21 @@ jobs:
...
+    name: Set-up and generate version name and version code
+    runs-on: ubuntu-22.04
+    outputs:
+      version-code: ${{ steps.version-code.outputs.version-code }}
+      version-name: ${{ steps.version-name.outputs.version-name }}
+
     steps:
       - name: Run checkout github action
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```
jonnyandrew, line 28: Why do we need to downgrade `checkout`?

We seem to be using inconsistent/outdated versions of the checkout action in some other jobs too. Please can we align all the versions?

.github/workflows/on_push_develop.yml
```diff
@@ -64,72 +58,263 @@ jobs:
...
+
+  style-checks:
+    name: Run style (detekt, lint & vale) checks
+    runs-on: ubuntu-22.04
+    needs:
+      - set-up
+    permissions:
+      # Set permissions for ${{ secrets.GITHUB_TOKEN }}
+      # https://docs.github.com/en/actions/security-guides/automatic-token-authentication
+      contents: read
```
jonnyandrew: I believe `contents: read` is granted by default, so we could remove

.github/workflows/on_push_develop.yml
```diff
@@ -64,72 +58,263 @@ jobs:
...
           test-type: "style"
           github-token: ${{ secrets.MODULE_FETCH_TOKEN }}
           github-actor: ${{ secrets.MODULE_FETCH_TOKEN_USERNAME }}
-          version-code: ${{ steps.version-code.outputs.version-code }}
-          version-name: "${{ steps.latest-tag.outputs.current-tag }}-${{ steps.version-code.outputs.version-code }}"
+          version-code: ${{ needs.set-up.outputs.version-code }}
+          version-name: ${{ needs.set-up.outputs.version-name }}
           app-check-token: ${{ secrets.BUILD_APPCHECK_DEBUG_TOKEN }}
 
+      - name: Upload ktlint coverage report
```
jonnyandrew: Is the name 'coverage report' correct here?

I think the ktlint, detekt and Android lint reports might be classed as a 'lint report', 'style report', or 'static analysis report' depending on the tool.

Does this impact the naming of the `mobile-android-pipelines/actions/upload-coverage-reports` and `mobile-android-pipelines/actions/download-coverage-reports` actions?

Perhaps we can rename:
- `upload-coverage-reports` -> `upload-reports`
- `restore-coverage-reports` -> `download-reports`

config/actions/sonar-analysis/action.yml
```diff
@@ -19,13 +19,36 @@ inputs:
...
+        find . -path "*/build/reports/detekt*" -name "*.xml" | sort
+        echo "=== Lint reports ==="
+        find . -path "*/build/reports/lint-results*" -name "*.xml" | sort
+        echo "=== Ktlint reports ==="
+        find . -path "*/build/reports/ktlint*" -name "*.xml" | sort
+
     - name: Run Sonarqube PR Analysis
       if: ${{ github.event_name == 'pull_request' }}
       run: |
+        ./gradlew createFullJarBuildDebugTestFixtures \
```
jonnyandrew: Could we generate all the Jars in one of the prerequisite jobs and then upload them for the sonar jobs to use?

As an example: https://github.com/govuk-one-login/mobile-android-cri-orchestrator/blob/18da06f7875e002845f126eef1ea8b219b8b94f5/.github/workflows/test.yml#L86-L92

I think that should help us avoid compiling code twice in sequence, saving time.

config/actions/upload-coverage-reports/action.yml
```diff
@@ -0,0 +1,19 @@
...
+  path:
+    description: "Path for the reports"
+    required: false
+    default: "**/reports/jacoco/**/*.xml"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Upload coverage reports
+      uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
```
jonnyandrew, line 16: This version is quite out of date - can we use the latest please?

README.md
jonnyandrew, line 1: It's really nice to see us being able to split up the tests into different jobs. However the work highlights that we duplicate our lint and test steps across 5 different GitHub Actions workflows.

Splitting what we already have into jobs increases both the amount and complexity of the workflow code and I'm concerned that this tips them over to being unmaintainable. And increases the risk of deviations between workflows and potential security issues going unnoticed.

Do you think we'd be able to consolidate this into a single `check.yml` workflow that can run on all of the various triggers (on PR, on push to develop/main, on schedule)? (Similar to how [mobile-android-cri-orchestrator/.github/workflows/test.yml](https://github.com/govuk-one-login/mobile-android-cri-orchestrator/blob/18da06f7875e002845f126eef1ea8b219b8b94f5/.github/workflows/test.yml#L4-L11) works)

If you agree, it's probably something we'd want to do prior to merging this PR.

.github/workflows/on_push_main.yml
```diff
@@ -110,60 +104,407 @@ jobs:
...
+      - name: Run sonar quality gate
+        uses: ./config/actions/sonar-quality-gate
+        with:
+          sonar-token: ${{ secrets.SONAR_TOKEN }}
+
+      - name: Clean workspace
+        uses: ./mobile-android-pipelines/actions/clean-workspace
+
+  build-and-publish-release-integration:
+    name: Verify code base when pushed on release
```
BiancaMihaila: Change name

config/actions/upload-coverage-reports/action.yml
BiancaMihaila: There's a PR here by dneepndabot  https://github.com/govuk-one-login/mobile-android-one-login-app/pull/818 - will do all of that after this is merged in, since it's easier

.github/workflows/run_checks.yml
BiancaMihaila: Done here: https://github.com/govuk-one-login/mobile-android-pipelines/pull/311 - I'll update once that is merged in

.github/workflows/on_push_develop.yml
BiancaMihaila: There's a [PR raised by dependabot to update all of them to the latest](https://github.com/govuk-one-login/mobile-android-one-login-app/pull/818), this is why I haven't touched it, I think the reason why this happened is because I copied some actions form different ones - you happy to let this this way and once this RP is merged I'll rebase the dependabot and merge that in so it will bring all of them on the same page?

.github/workflows/on_push_develop.yml
BiancaMihaila: I think it might be good to keep it so it's a bit more explicit since it won't make much of a difference if it's applied by default as well?

.github/workflows/on_push_develop.yml
BiancaMihaila: From m point of view, I like the idea of keeping coverage so it's more explicit is used for the code coverage analysis in sonar
I think it can be done, but I don't think it's required for this, maybe as an enhancement since the change would be in the pipelines repo

.github/workflows/run_checks.yml
BiancaMihaila: All done

README.md
BiancaMihaila: Done all this - teh flows are the same because we always will run those and in that order, but the actual set-up is having only one run checks workflow that gets triggered and is completely separated from the publishing etc, which reduced a lot of duplication

.github/workflows/on_push_release.yml
```diff
@@ -2,6 +2,10 @@
 name: On Branch Push (release)
 
 on:
+  workflow_run:
+    workflows: [ Run checks (tests and style checks) ]
+    branches: [ release/** ]
+    types: [ completed ]
   push:
     branches:
       - "release/**"
```
jonnyandrew: _This comment would also apply to the `on_push_main` and `on_push_develop` workflows_

After this change, the workflow will be triggered in both of the following scenarios:
- after the check workflow is complete
- immediately on push

Could that result in the workflow running twice?

I assume we still need the dependency submission job to run on push. But we want the deployment to run after checks complete.

As a suggestion:
- deployment workflow(s) are triggered after the `run_checks` workflow
- a separate dependency submission workflow is triggered on push to any branch

.github/workflows/on_push_release.yml
```diff
@@ -11,143 +15,124 @@ concurrency:
...
-  prepare-for-release:
-    name: Verify code base pre-release
+  dependency-submission-job:
+    name: Generate and submit dependency graph
     runs-on: ubuntu-22.04
-    #    runs-on: ubuntu-20.04-16core # Larger github runner, with KVM acceleration
     permissions:
-      # Set permissions for ${{ secrets.GITHUB_TOKEN }}
-      # https://docs.github.com/en/actions/security-guides/automatic-token-authentication
+      contents: write
```
jonnyandrew: Does dependency submission need `contents: write`? I thought that's for creating a release or pushing a tag for example?

.github/workflows/on_push_release.yml
```diff
@@ -11,143 +15,124 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  id-token: write
```
jonnyandrew: We're using the `id-token: write` permission in a few places but I'm not sure it's obvious what we're using it for. Perhaps we can add a tiny code comment above each usage to explain what it's for?

.github/workflows/run_checks.yml
```diff
@@ -1,9 +1,18 @@
...
+      - opened
+      - reopened
+      - synchronize
+  push:
+    branches:
+      - main
+      - develop
   schedule:
     - cron: 0 3 * * *
+  merge_group:
```
jonnyandrew: Is this needed? If I understood the docs correctly it's only relevant for repos using merge queues.

.github/workflows/run_checks.yml
```diff
@@ -350,9 +351,13 @@ jobs:
         uses: ./mobile-android-pipelines/actions/bundle-reports
 
       - name: Run sonar quality gate
+        id: sonar-gate
         uses: ./config/actions/sonar-quality-gate
         with:
           sonar-token: ${{ secrets.SONAR_TOKEN }}
 
+      - name: Set status output
+        run: echo "status=success" >> $GITHUB_OUTPUT
```
jonnyandrew: I don't think the `GITHUB_OUTPUT` environment will be available as you might expect between workflows. To share data between workflows you'll need to upload/download artifacts unfortunately.

Having said that, I can't see this output being used anywhere - can we remove it?

.github/workflows/run_checks.yml
```diff
@@ -1,9 +1,18 @@
...
+      - reopened
+      - synchronize
+  push:
+    branches:
+      - main
+      - develop
   schedule:
     - cron: 0 3 * * *
+  merge_group:
   workflow_dispatch:
```
jonnyandrew: I wonder if we could remove `workflow_dispatch` unless needed for a specific use case?

.github/workflows/on_push_release.yml
```diff
@@ -2,6 +2,10 @@
 name: On Branch Push (release)
 
 on:
+  workflow_run:
+    workflows: [ Run checks (tests and style checks) ]
+    branches: [ release/** ]
+    types: [ completed ]
```
jonnyandrew: Note that the `completed` type includes failed workflows. Was the intention to run on success?

If so we can use a condition:

```
if: ${{ github.event.workflow_run.conclusion == 'success' }}
```

config/actions/sonar-analysis/action.yml
```diff
@@ -43,17 +63,26 @@ runs:
         INPUT_VERSION_CODE: ${{ inputs.version-code }}
         INPUT_VERSION_NAME: ${{ inputs.version-name }}
         BUILD_DEBUG_APP_CHECK_TOKEN: ${{ inputs.app-check-token }}
+        GITHUB_ACTOR: ${{ inputs.github-actor }}
+        GITHUB_TOKEN: ${{ inputs.github-token }}
 
     - name: Run Sonarqube Branch Analysis
       if: ${{ github.event_name != 'pull_request' }}
       run: |
+        ./gradlew createFullJarBuildDebugTestFixtures \
```
jonnyandrew: I saw that you were able to remove this task from the action's PR analysis step.

Are we able to do the same here?

- related: https://github.com/govuk-one-login/mobile-android-one-login-app/pull/819#discussion_r3118523986

README.md
```diff
@@ -23,6 +23,88 @@ This project uses a git-flow branching strategy, briefly:
 - The new tag will be pushed on the merge to `main` with the default merge title
 - `main` should then be merged back into `develop`
 
+## Pull Request workflow
```
jonnyandrew, line 26: I think this documentation may be out of date after the most recent changes.

Given that GitHub visualises workflows for us, do we need to keep a detailed diagram of the job breakdown in the readme? Perhaps we could make this slightly higher level so that it's less likely to fall out of date over t.me

.github/workflows/on_push_release.yml
BiancaMihaila: I think on merge_to_main we probably don't want that because it delays the release equity a bit - I think generally all the checks ran on PR are the same on-merge so maybe if that changes in the future we should probably update accordingly if we are not confident to publish directly 
On release generally the Tech Lead won't approve until the tests are ran so we can remove that from there
On push develop I've added the workflow run checks as a pre-requisite before publishing as we don't usually have time pressure/ constraints  - so we can only keep it there
But happy to talk about this if it helps and we can maybe decide on how to go forward 
I'll make the changes accordingly to not have those running twice

.github/workflows/on_push_release.yml
BiancaMihaila: Not fully sure, I think you're right and just got kept in, I'll change now and test it when it runs on PR
It might be needed to update the outcomes (security vulnerabilities found, etc) but not 100% because I'm not sure how exactly it works but I'll try it now

.github/workflows/on_push_release.yml
jonnyandrew: That makes sense. So we can just remove the `workflow_run` triggers and continue running these on push?

.github/workflows/run_checks.yml
BiancaMihaila: This was a leftover from when I was debugging to make the sonar coverage work - thank you, I'll remove now

.github/workflows/on_push_develop.yml
jonnyandrew: My rationale for removing would be to reduce cognitive load when reading this workflow.

But it really is a nit-pick, definitely not blocking.

.github/workflows/on_push_develop.yml
jonnyandrew: Maybe I misunderstood, as I thought 'coverage' usually only refers to test coverage? 

Regardless, it's minor and I agree it can be improved later.

README.md
BiancaMihaila: I think it's still accurate, I made this to be a bit more of a visual representation of what actually happens rather than the equivalent of the actual workflow
If we make it more high-level won't exactly show what checks we're running, the only change to do that would be to just mention running checks? If we're happy with that I can amend it, so if any checks get removed/ added then we won't need to amend this as well?

.github/workflows/on_push_release.yml
BiancaMihaila: I think I might move teh dependency submission into it's own workflow to run independently, seems easier to maintain and remove duplication, as we are using the same one in 2 places at the moment (and I think I mixed and put it on merge main instead on pull_request - so that make it a bit more clear)

config/actions/sonar-analysis/action.yml
BiancaMihaila: Done

.github/workflows/on_push_release.yml
BiancaMihaila: I've amended it but need to test it works as expected - and we need to merge this to check that because PR workflow don't have permission to start the publishing

.github/workflows/run_checks.yml
BiancaMihaila: Done, wasn't needed

.github/workflows/run_checks.yml
BiancaMihaila: Done, thank you

.github/workflows/on_push_release.yml
BiancaMihaila: Done, thank you!

README.md
jonnyandrew: I'd commented this under the assumption we were using workflow chaining so by high level I meant to show that.

Given it's still accurate, I'm happy to leave as-is - sorry for the confusion

.github/workflows/on_push_develop.yml
BiancaMihaila: Makes sense - thank you!

.github/workflows/on_push_release.yml
BiancaMihaila: It seems to be required to actually update the dependency graph after the analysis is done so and to add it back in

README.md
BiancaMihaila: Thank you 🙏

.github/workflows/on_push_release.yml
jonnyandrew: `on_push_release.yml` still seems to include both triggers - are we planning to remove `workflow_run` here too?

BiancaMihaila: I think I removed it all

.github/workflows/run_checks.yml
```diff
@@ -0,0 +1,358 @@
...
+    name: Run sonar analysis and check sonar quality gate
+    runs-on: ubuntu-22.04
+    outputs:
+      status: ${{ steps.sonar-gate.outputs.status }}
+    needs:
+      - set-up
+      - style-checks
+      - unit-checks
+      - component-checks
+      - snapshot-checks
```
alex-bradbury-gds: Is this needed for sonar checks?

BiancaMihaila: Removed the snapshot checks from there as you're completely right - thank you 🙌

alex-bradbury-gds (approved):

PR #825, opened 2026-04-27
feat: Add modifier that configures different behaviour for tab and arrow up/ down

BiancaMihaila:
- implement this on the `SettingsScreen`

**Evidence**

**Tab**

[Screen_recording_20260429_091215.webm](https://github.com/user-attachments/assets/54c2e1fd-a104-4ae4-8ccd-17a6ec91b4a0)

**Arrow up/ down**

[Screen_recording_20260429_091346.webm](https://github.com/user-attachments/assets/58304908-742c-45a7-9146-cd4db99164cd)

Resolves: DCMAW-19031


features/src/main/java/uk/gov/onelogin/features/settings/ui/SettingsScreen.kt
```diff
@@ -597,6 +612,49 @@ internal fun SettingsScreenOptInNoShowBiometricsPreview() {
...
 }
 
+/**
+ * Adds a downwards and upwards scroll when a keyboard down or up arrow is pressed
+ *
+ * @param scrollState [ScrollState] represents the column scroll state
+ * @return augmented [Modifier]
+ */
+@Composable
+fun Modifier.bringIntoView(scrollState: ScrollState): Modifier {
```
jonnyandrew: We have this util function available in the design system library. Can we reuse that?

https://github.com/govuk-one-login/mobile-android-ui/blob/1940c835277038814e8267941822a389559f3ad8/patterns/src/main/java/uk/gov/android/ui/patterns/leftalignedscreen/LeftAlignedScreenContentV2.kt#L363

BiancaMihaila: That one is applied to a LazyListState not a Column ScrollState - so unfortunately it's not compatible with the SettingsScreen set-up

jonnyandrew: Could we update it to support both? For example: https://github.com/govuk-one-login/mobile-android-ui/pull/421

jonnyandrew (approved):
Thanks!

PR #755, opened 2026-02-24
docs: Add diagram for HandleLocalLoginImpl

BiancaMihaila:
## Context

DCMAW-19841

PR #837, opened 2026-05-04
refactor: DCMAW-19840 Extract remote login logic from view model to use case

BiancaMihaila:
Notes from the original author
> - move logic to start login and handle the errors or successful
redirect to the app
> - move logic once the redirect is successful and user can perform token
request
> - move logic handling the TokenResponse and when/ what to save to open
secure store, token secure store or just in memeory
> - handle local auth in the same use case

## Context

This PR extracts the remote login logic from the presentation layer of the welcome screen (sign-in screen) into a dedicated login view model (`LoginViewModel`), and remote login domain use case (`RemoteLoginImpl`).

DCMAW-19840

## Out of scope

- Suspending callback based APIs https://github.com/govuk-one-login/mobile-android-one-login-app/pull/1032


features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,269 @@
...
+            }
+        }
+
+        override suspend fun finalise(
+            intent: Intent,
+            isReAuth: Boolean,
+            activity: FragmentActivity,
+        ) {
+            // Enables running all functions in respective order as this cannot be done async
+            // and allows to use teh same Coroutine scope avoiding multithreading
```
alex-bradbury-gds:
```suggestion
            // and allows to use the same Coroutine scope avoiding multithreading
```
;)

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,269 @@
...
+                    deferred
+                        .complete(
+                            Result.failure(
+                                it
+                                    ?: Exception("Login finalise failed")
+                            )
+                        )
+                }
+            )
+            coroutineScope {
```
alex-bradbury-gds:
```suggestion
            withContext(Dispatchers.IO) {
```
Wonder if we should be specifying the dispatcher given it's doing network operations... 🤔

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,269 @@
...
+                            it.message.toString(),
+                            loginException,
+                        )
+                        handleLoginErrors(it)
+                    }
+                )
+            }
+        }
+
+        private fun CoroutineScope.handleTokens(
```
alex-bradbury-gds: Is there a reason for this being an extension on `CoroutineScope` and not just a `private suspend fun...`?

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,269 @@
...
+            val jwksUrl =
+                context.getString(
+                    R.string.stsUrl,
+                    context.getString(R.string.jwksEndpoint),
+                )
+            this.launch {
+                if (!verifyIdToken(tokens.idToken, jwksUrl)) {
+                    navigator.navigate(LoginRoutes.SignInRecoverableError, true)
+                } else {
+                    withContext(Dispatchers.Main) {
```
alex-bradbury-gds: Do you actually need this here?

features/src/main/java/uk/gov/onelogin/features/login/ui/signin/welcome/WelcomeScreen.kt
```diff
@@ -79,8 +79,8 @@ fun WelcomeScreen(
     LifecycleEventEffect(Lifecycle.Event.ON_START) {
         if (!loading) {
             analyticsViewModel.trackWelcomeView()
+            viewModel.stopLoading()
```
alex-bradbury-gds: why move this?

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
BiancaMihaila: yes, to allow to be called from within a non-suspend lambda - otherwise line 105 and 113 won't work, it would need flags and to be moved outside of the lambdas and that is a bit messy

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
BiancaMihaila: We might be able to remove this actually, I think I forgot to after I fixed a small concurrency bug that popped up - I'll test that now

features/src/main/java/uk/gov/onelogin/features/login/ui/signin/welcome/WelcomeScreen.kt
BiancaMihaila: that was again something I forgot to change back when I was debugging  🤦‍♀️ - I'll move it back

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
BiancaMihaila: Sadly , not possible because the handleTokens function required the LocalAuthManager prompt at some point, which requires the Main thread to display the UI - but I can change it to `withContext(Dispatchers.Main)` instead which is better than the `coroutineScope`

app/src/main/java/uk/gov/onelogin/login/LoginUseCaseModule.kt
```diff
@@ -23,8 +39,48 @@ interface LoginUseCaseModule {
...
-    fun bindHandleLoginRedirect(handleRedirect: HandleLoginRedirectImpl): HandleLoginRedirect
+    fun bindHandleLoginRedirect(handleRedirect: FinaliseRemoteLoginImpl): FinaliseRemoteLogin
+}
+
+@InstallIn(ViewModelComponent::class)
+@Module
+object LoginUseCaseObjectModule {
+    @Suppress("LongParameterList")
+    @Provides
+    fun bindRemoteLogin(
```
jonnyandrew: Could this be `@Binds` rather than `@Provides`?

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/finalise/FinaliseRemoteLogin.kt
```diff
@@ -1,9 +1,12 @@
-package uk.gov.onelogin.features.login.domain.signin.loginredirect
+package uk.gov.onelogin.features.login.domain.signin.remotelogin.finalise
 
 import android.content.Intent
 import uk.gov.android.authentication.login.TokenResponse
 
-fun interface HandleLoginRedirect {
+/**
+ * Use case allowing to perform login (first time user) and
```
jonnyandrew: This sentence seems to be incomplete

features/src/main/java/uk/gov/onelogin/features/login/LoginViewModel.kt
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+        }
+
+        fun abortLogin(
+            launcher: ActivityResultLauncher<Intent>,
+            isReAuth: Boolean
+        ) {
+            _loading.value = false
+            startLoginActivity(launcher, isReAuth).cancel()
```
jonnyandrew: Why is `abortLogin()` calling `startLoginActivity()`? I find this a bit confusing and wonder if it's doing what we expect.

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,266 @@
+package uk.gov.onelogin.features.login.domain.signin.remotelogin
+
+import android.content.Context
+import android.content.Intent
+import androidx.activity.result.ActivityResultLauncher
+import androidx.fragment.app.FragmentActivity
+import dagger.hilt.android.qualifiers.ApplicationContext
+import dev.zacsweers.metro.Inject
```
jonnyandrew: Is this supposed to use Metro `@Inject`?

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,266 @@
...
+                                it
+                                    ?: Exception("Login finalise failed")
+                            )
+                        )
+                }
+            )
+            // Required to ensure the navigation and errors are handled accordingly - should look into simplifying this or if possible to
+            // refactor and not use the Main thread
+            withContext(Dispatchers.Main) {
+                deferred.await().fold(
```
jonnyandrew: We seem to have a number of callback based APIs creating complexity in this codebase.

I think it would be worth looking at refactoring those to suspend functions that return a result type.

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,266 @@
...
+                            Result.failure(
+                                it
+                                    ?: Exception("Login finalise failed")
+                            )
+                        )
+                }
+            )
+            // Required to ensure the navigation and errors are handled accordingly - should look into simplifying this or if possible to
+            // refactor and not use the Main thread
+            withContext(Dispatchers.Main) {
```
jonnyandrew: Could we inject the main dispatcher to use here?

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,266 @@
...
+
+            localAuthManager.enforceAndSet(
+                false,
+                activity = activity,
+                callbackHandler =
+                    object : LocalAuthManagerCallbackHandler {
+                        override fun onSuccess(backButtonPressed: Boolean) {
+                            // Required to enable running suspend functions form within the callback handler
+                            // Requires Main thread as it displays UI that needs to be persisted until user allows for continuation of the flow to the other screens
+                            CoroutineScope(Dispatchers.Main).launch {
```
jonnyandrew: If this is needed, please can we inject the coroutine scope as shown in https://developer.android.com/kotlin/coroutines/coroutines-best-practices#create-coroutines-data-layer

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,266 @@
...
+        private val tokenRepository: TokenRepository,
+        private val verifyIdToken: VerifyIdToken,
+        private val autoInitialiseSecureStore: AutoInitialiseSecureStore,
+        private val savePersistentId: SavePersistentId,
+        private val saveTokenExpiry: SaveTokenExpiry,
+        private val signOutUseCase: SignOutUseCase,
+        private val removeRefreshTokenAndExpiry: RemoveRefreshTokenAndExpiry,
+        private val errorCounter: Counter,
+        private val logger: Logger,
+        private val navigator: Navigator
```
jonnyandrew: This class is responsible for lots of different navigation.

I imagine that the view models that call this use case are also responsible for different navigation.

Would it be worth consolidating the navigation logic in the view models? We could return different results from `RemoteLogin` and let the view model decide where to navigate to based on that?

jonnyandrew: It turns out this would be non-trivial because we have several layers of suspending callback based APIs. So on second thought, I think we ought to address that problem first.

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
jonnyandrew: I will fix this in a follow-up PR

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
jonnyandrew: This doesn't appear to be needed. The navigator already switches work to the main thread internally.

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
jonnyandrew: This doesn't appear to be needed. The navigator already switches work to the main thread internally.

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,261 @@
...
+            autoInitialiseSecureStore.initialise(tokens.refreshToken)
+            if (tokens.refreshToken != null) {
+                if (pref is LocalAuthPreference.Enabled) {
+                    saveRefreshTokenExpiryToOpenStore(tokens)
+                }
+            } else {
+                removeRefreshTokenAndExpiry.remove()
+            }
+            navigator.navigate(MainNavRoutes.Start, true)
+        }
```
jonnyandrew: I'm not sure I understand why the logic is different between these functions.

For a re-auth, `autoInitialiseSecureStore.initialise(refreshToken)` is only called if local auth is enabled.

For a first time user, `autoInitialiseSecureStore.initialise(refreshToken)` is always called irrespective of the local auth settings.

___ 
Edit: [It turns out that `AutoInitialiseSecureStore.initialise(refreshToken)` checks the local auth status internally. If local auth is disabled (or unset) the function is a no-op.](https://github.com/govuk-one-login/mobile-android-one-login-app/blob/661e698db4f5ff1a96bb463c1eb784a8780ea9a9/core/src/main/java/uk/gov/onelogin/core/tokens/data/initialise/AutoInitialiseSecureStoreImpl.kt#L37)

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,261 @@
...
+        private val autoInitialiseSecureStore: AutoInitialiseSecureStore,
+        private val savePersistentId: SavePersistentId,
+        private val saveTokenExpiry: SaveTokenExpiry,
+        private val signOutUseCase: SignOutUseCase,
+        private val removeRefreshTokenAndExpiry: RemoveRefreshTokenAndExpiry,
+        private val errorCounter: Counter,
+        private val logger: Logger,
+        private val navigator: Navigator
+    ) : RemoteLogin {
+        override suspend fun start(launcher: ActivityResultLauncher<Intent>) {
```
jonnyandrew: TODO: This function isn't covered by unit tests

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,261 @@
...
+                        },
+                        onFailure = {
+                            val exception = it ?: Exception("Login finalise failed")
+                            deferred.complete(Result.failure(exception))
+                        }
+                    )
+                }
+                .await()
+                .onSuccess {
+                    handleTokens(it, isReAuth, activity)
```
jonnyandrew: TODO: This function calls multiple levels of private functions - flatten?

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,261 @@
...
+                    logger.error(
+                        loginException.javaClass.simpleName,
+                        it.message.toString(),
+                        loginException,
+                    )
+                    handleLoginErrors(it)
+                }
+        }
+
+        private suspend fun handleTokens(
```
jonnyandrew: TODO: return the navigation route

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,261 @@
...
+        private suspend fun saveAccessTokenExpiryToOpenStore(tokens: TokenResponse) {
+            saveTokenExpiry.saveExp(
+                ExpiryInfo(
+                    key = ACCESS_TOKEN_EXPIRY_KEY,
+                    value = tokens.accessTokenExpirationTime,
+                ),
+            )
+        }
+
+        private suspend fun handleLoginErrors(it: Throwable?) {
```
jonnyandrew: TODO: Return the navigation route / action

features/src/main/java/uk/gov/onelogin/features/login/ui/signin/welcome/WelcomeScreen.kt
```diff
@@ -80,37 +84,27 @@ fun WelcomeScreen(
...
+        loginViewModel.stopLoading()
     }
 }
 
 private fun handleResult(
     result: ActivityResult,
-    viewModel: WelcomeScreenViewModel,
+    loginViewModel: LoginViewModel,
     context: FragmentActivity,
 ) {
```
jonnyandrew: Not sure this function is needed any more

features/src/main/java/uk/gov/onelogin/features/login/LoginViewModel.kt
```diff
@@ -0,0 +1,125 @@
...
+            launcher: ActivityResultLauncher<Intent>,
+            isReAuth: Boolean
+        ) = viewModelScope.launch {
+            _loading.emit(true)
+            if (onlineChecker.isOnline()) {
+                // This is checking for the re-auth path - if persistent session ID is empty on re-auth, that suggests
+                // something went wrong and we delete all data - otherwise, we allow the user to do a re-auth
+                if (getPersistentId().isNullOrEmpty() && isReAuth) {
+                    try {
+                        signOutUseCase.invoke()
```
jonnyandrew: - [ ] Ensure complete test coverage of LoginViewModel
- [ ] Refactor so this isn't deeply nested

features/src/testFixtures/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/TestRemoteLogin.kt
```diff
@@ -0,0 +1,46 @@
...
+import androidx.fragment.app.FragmentActivity
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.awaitCancellation
+
+class TestRemoteLogin : RemoteLogin {
+    var started = false
+        private set
+    var startCancelled = false
+        private set
+    var finalisedWith: FinalArgs? = null
```
jonnyandrew: Perhaps this could be left up to a mockito spy()

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
jonnyandrew: https://github.com/govuk-one-login/mobile-android-one-login-app/pull/837/commits/dbcdf6094362e7d73c2813160385e2c36c953172

.sh/tests/runInstrumentationChecks
```diff
@@ -18,7 +18,7 @@ set -o errexit
 # Provide a default value for WORKER_COUNT
 if [[ -z "${WORKER_COUNT}" ]]
 then
-  WORKER_COUNT=2
+  WORKER_COUNT=1
```
jonnyandrew: This is a temporary fix that we can revert after merging
- https://github.com/govuk-one-login/mobile-android-one-login-app/pull/1018

README.md
```diff
@@ -159,3 +159,126 @@ You can use the following command to check the SHA 256 checksum of a file
 ```bash
 shasum -a 256 gradle-8.10.2-bin.zip
 ```
+
+## Flow Diagrams
```
jonnyandrew: TODO: Split these diagrams to a separate PR?

jonnyandrew:
I have kept the overall structure from the original PR, but refactored a few things so that I could understand the logic we were implementing.

My changes are in the [most recent commits](https://github.com/govuk-one-login/mobile-android-one-login-app/pull/837/changes/5a6fa5d6b8a7de0b703d82a8c96a5452a324e1c8..05fe04627fcc429921b618d5be0740a5e19dcca3#diff-365a2982dfca9cbd494863085fe809931591fc1759c55fb1494001aeaed3da4d). It should also be possible to review each of my commits on its own.

README.md
jonnyandrew: Should be covered in https://github.com/govuk-one-login/mobile-android-one-login-app/pull/755

.sh/tests/runInstrumentationChecks
jonnyandrew: https://github.com/govuk-one-login/mobile-android-one-login-app/pull/837/commits/2472084368e1a7fad1f2ab31aba534802fef8a9b

features/src/main/java/uk/gov/onelogin/features/login/ui/signin/welcome/WelcomeScreen.kt
```diff
@@ -41,31 +40,39 @@ import uk.gov.onelogin.core.ui.pages.EdgeToEdgePage
...
 
 @Composable
 fun WelcomeScreen(
     viewModel: WelcomeScreenViewModel = hiltViewModel(),
+    loginViewModel: LoginViewModel = hiltViewModel(),
     analyticsViewModel: SignInAnalyticsViewModel = hiltViewModel(),
     loadingAnalyticsViewModel: LoadingScreenAnalyticsViewModel = hiltViewModel(),
 ) {
-    val loading by viewModel.loading.collectAsState()
+    val loading by loginViewModel.loading.collectAsState()
```
elijahhezekiah-maker, line 52: Do we want to consider extending this into a UIstate  using data class / sealed  interface that can gives us a clear model of the behaviour on the welcome screen. 

This will be similar to the WalletViewModel Uistate. 

I am open to reasons why that will not be a good idea

jonnyandrew: I haven't made any changes to how the screen state is represented here because there are still only two screen states that the existing boolean captures.

A screen state class can be useful if we have a group of different state that are shown together e.g.

```kt
data class ScreenState(
   val displayContinueButton: Boolean,
   val warning: String?,
)
```

Sealed classes are useful if the screen has multiple modes that are mutually exclusive e.g. loading, display and error modes:

```kt
sealed interface ScreenState {
    data object Loading: ScreenState
    data class Display(content: String): ScreenState
    data class Error(message: String): ScreenState
}
```

Having said that, there are also reasons that we might want to create the abstraction before it's needed:
1. It's a convention in the codebase
2. We expect this screen's state to be extended (and want to put the framework in place for that)

But as neither of these apply, I think we can safely leave it as `val display: Boolean`.

elijahhezekiah-maker: We could  group all of the scenarios that return an error  into a unified  failure state and then handle the success state as well.  

that way  when someone reads the UI Layer, they can focus on the high-level relationship between  the view and  ViewModel  first (Loading vs. Success vs. Failure) without getting overwhelmed by specific  functions and error details right away 

 I think there might be enough shared commonalities  to justify  the abstraction 

```

sealed interface UiState {
   
    object Loading : LoginUiState
    data class Success(val isReAuth: Boolean) : LoginUiState
   
    sealed interface Failure : LoginUiState 
}

```

but if you feel that the abstraction is not needed then that's fine.

elijahhezekiah-maker (approved):

features/src/main/java/uk/gov/onelogin/features/login/domain/signin/remotelogin/RemoteLoginImpl.kt
```diff
@@ -0,0 +1,253 @@
...
+                        loginException,
+                    )
+
+                    if (it is AuthenticationError && it.type == AuthenticationError.ErrorType.SERVER_ERROR) {
+                        recoverableErrors.increment()
+                    }
+
+                    val result = it.toLoginResult()
+
+                    if (result is RemoteLogin.Result.Failure && result.type == RemoteLogin.FailureType.AccessDenied) {
```
alex-bradbury-gds, line 98: Maybe some comment here just to explain what we're doing and why we're doing it

jonnyandrew: I've added a comment here, linking to the tech design: https://github.com/govuk-one-login/mobile-android-one-login-app/pull/837/commits/75af2bdb1fc60eff1b6be50d0d951ff8cca3912e

features/src/main/java/uk/gov/onelogin/features/login/ui/signin/welcome/WelcomeScreenViewModel.kt
alex-bradbury-gds, line 1: -266, dream scenario 😆 👌

alex-bradbury-gds (approved):

PR #847, opened 2026-05-07
ci: Add lint for GitHub Actions workflows

jonnyandrew:
## Changes

Add lint action for GitHub Actions workflows.

## Evidence

- https://github.com/govuk-one-login/mobile-android-one-login-app/pull/847/commits/bfb1ebb68bb65afa82dccb848faec387fb1662d6: [Job fails](https://github.com/govuk-one-login/mobile-android-one-login-app/actions/runs/25486764467/job/74784382498?pr=847) due to problems with the workflow.
- https://github.com/govuk-one-login/mobile-android-one-login-app/pull/847/commits/8a2d946992b917d160e247ac1be486cbb0d59897: [Job passes](https://github.com/govuk-one-login/mobile-android-one-login-app/actions/runs/25486982902/job/74785170796?pr=847) after fixes

## Context

DCMAW-20179

Helps prevent workflow regressions such as broken references in GitHub Actions.

Follows the same pattern as [mobile-android-cri-orchestrator/blob/main/.github/workflows/lint.yml](https://github.com/govuk-one-login/mobile-android-cri-orchestrator/blob/main/.github/workflows/lint.yml#L74-L83) which also uses [alphagov/govuk-infrastructure](https://github.com/alphagov/govuk-infrastructure)'s GitHub action.

### DIFF
--- a/.github/workflows/run_checks.yml
+++ b/.github/workflows/run_checks.yml
@@ -297,8 +297,6 @@ jobs:
   sonar-analysis:
     name: Run sonar analysis and check sonar quality gate
     runs-on: ubuntu-22.04
-    outputs:
-      status: ${{ steps.sonar-gate.outputs.status }}
     needs:
       - set-up
       - style-checks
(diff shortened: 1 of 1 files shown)


.github/workflows/run_checks.yml
```diff
@@ -297,8 +297,6 @@ jobs:
   sonar-analysis:
     name: Run sonar analysis and check sonar quality gate
     runs-on: ubuntu-22.04
-    outputs:
-      status: ${{ steps.sonar-gate.outputs.status }}
```
elijahhezekiah-maker, line 301: @jonnyandrew can you confirm why we are no longer sending the  sonar output in this workflow?

jonnyandrew: The `steps.sonar-gate.outputs.status` doesn't exist so this job output would be empty.

Also, no job is dependent on the sonar analysis job, so the whole job output is redundant.

image

elijahhezekiah-maker (approved):

PR #849, opened 2026-05-07
feat: enhance logs for AppInfo errors returned for /appInfo API calls

BiancaMihaila:


features/src/main/java/uk/gov/onelogin/features/appinfo/data/model/AppInfoRemoteState.kt
```diff
@@ -7,6 +7,7 @@ sealed class AppInfoRemoteState {
 
     data class Failure(
         val reason: String,
+        val error: Throwable? = null,
```
jonnyandrew: Does this need to be nullable?

It looks like the error that we're supplying in `AppInfoRemoteSourceImpl` is non-null.

features/src/main/java/uk/gov/onelogin/features/appinfo/domain/AppInfoServiceImpl.kt
```diff
@@ -25,11 +26,15 @@ class AppInfoServiceImpl
                 }
                 AppInfoRemoteState.Offline -> useLocalSource(AppInfoServiceState.Offline)
                 is AppInfoRemoteState.Failure -> {
-                    val apiException = ApiInfoException(Exception(remoteResult.reason))
+                    val apiException =
+                        remoteResult.error?.let {
+                            ApiInfoException(it)
+                        } ?: ApiInfoException(Exception(remoteResult.reason))
```
jonnyandrew: If `remoteResult.error` can be made non-null (see previous comment), this could be simplified.

features/src/main/java/uk/gov/onelogin/features/appinfo/domain/AppInfoServiceImpl.kt
```diff
@@ -50,4 +55,9 @@ class AppInfoServiceImpl
                 fallback
             }
         }
+
+        companion object {
+            internal const val APP_INFO_ERROR_KEY = "AppInfo"
+            internal const val APP_INFO_REMOTE_ERROR_VALUE = "Remote app info failure"
```
jonnyandrew: Seeing as the keys are for filtering different errors, would it make sense that the they are kept to a relatively small set, and the values are the thing that vary?

So for example this error might have two custom keys:

- `component`/`feature` = `App Info`
- `action`/`scenario` = `Remote app info failure`

Some other points worth thinking about (outside of this PR):
- Should we be centralising some of these keys in mobile-android-logging?
- Do we need to agree some kind of namespacing system for domain specific keys (e.g. `wallet`/`idcheck`).

Edit: considering https://github.com/govuk-one-login/mobile-android-one-login-app/pull/849#discussion_r3246556544, I think we need to manage this in logging.

features/src/main/java/uk/gov/onelogin/features/appinfo/domain/AppInfoServiceImpl.kt
```diff
@@ -25,11 +26,12 @@ class AppInfoServiceImpl
                 }
                 AppInfoRemoteState.Offline -> useLocalSource(AppInfoServiceState.Offline)
                 is AppInfoRemoteState.Failure -> {
-                    val apiException = ApiInfoException(Exception(remoteResult.reason))
+                    val apiException = ApiInfoException(remoteResult.error)
                     logger.error(
                         apiException::class.simpleName.toString(),
                         remoteResult.reason,
                         apiException,
+                        ErrorKeys.StringKey(APP_INFO_ERROR_KEY, APP_INFO_REMOTE_ERROR_VALUE)
```
jonnyandrew: I think we need to reconsider how we're using custom keys for non-fatal exceptions.

As we have it currently, this key will be applied for every non-fatal exception tracked from this point in the flow onwards, meaning unrelated errors will be tracked as a "Remote app info failure".

This will pollute the data and make it difficult to filter the non-fatals in crashlytics.

jonnyandrew: Will be fixed in https://github.com/govuk-one-login/mobile-android-one-login-app/pull/868

features/src/main/java/uk/gov/onelogin/features/appinfo/domain/AppInfoServiceImpl.kt
jonnyandrew: Logger V2 doesn't support multiple custom keys. To resolve this comment we'll need to update to Logger V3: https://govukverify.atlassian.net/browse/DCMAW-20625

jonnyandrew:
Closing in favour of #874

PR #860, opened 2026-05-18
fix: Screenshots test not working

elijahhezekiah-maker:
- removed errorbody2 in AppIntegrityErrorScreen.kt
- testing CI with changes
- the task and CI Job should fail

After making the Changes to  the test-config.gradle.kts files, The snapshot test is failing as expected :

image

Resolves:[ DCMAW-20259](https://govukverify.atlassian.net/browse/DCMAW-20259)


build-logic/convention/src/main/java/uk/gov/onelogin/plugin/test-config.gradle.kts
```diff
@@ -22,6 +22,20 @@ open class TestTypeExtension(objects: ObjectFactory) {
 val extension = TestTypeExtension(project.objects)
 project.extensions.add("testTypeConfig", extension)
 
+/**
+ * Excludes the given test type's package from the original unit test task
+ * so tests aren't run twice (once in the original and once in the new task).
+ * "unit" and "snapshot" types are kept in the original task for Paparazzi verify/record.
+ */
+fun Project.excludeFromOriginalTask(testType: String, originalTestTaskName: String) {
+    val nonExcludableTypes = listOf("unit", "snapshot")
```
jonnyandrew: `snapshot` isn't in the set of possible test types so including snapshot in this list won't have any effect.

https://github.com/govuk-one-login/mobile-android-one-login-app/blob/a6f41ff3e46c1beb7d0ff6ad15d1c42f84123a0b/features/build.gradle.kts#L18-L25

build-logic/convention/src/main/java/uk/gov/onelogin/plugin/test-config.gradle.kts
```diff
@@ -43,11 +57,7 @@ project.afterEvaluate {
             val taskName = "${testType}Test${variantName}"
             val isDebug = buildType.name == "debug"
 
-            // Exclude this test type's package from the original unit test task
-            // so tests aren't run twice (once in the original and once in the new task)
-            tasks.named(originalTestTaskName, Test::class.java) {
-                exclude("**/$testType/**")
-            }
+            project.excludeFromOriginalTask(testType, originalTestTaskName)
```
jonnyandrew: It looks like previously we were excluding tests belonging to the new test tasks (e.g. `componentTestBuildDebug`) from the original test task (e.g. `testBuildDebugUnitTest`).

I'm not sure I agree with this approach because it doesn't align with what developers expect to happen when they run `testDebugUnitTest`, which is that  all the local JVM tests will run.

Given the exclusions also seems to interfere with the screenshot tests, what do you think about removing the exclusion behaviour entirely?

In doing this, I believe we'll also resolve [DCMAW-20329](https://govukverify.atlassian.net/browse/DCMAW-20329).

Concerning the original comment ("so tests aren't run twice"), I think we can address this if needed by limiting the tasks we're running.

build-logic/convention/src/main/java/uk/gov/onelogin/plugin/test-config.gradle.kts
jonnyandrew: For my own understanding, why was this exclusion interfering with the screenshot tests?

jonnyandrew (commented):
Well done for getting to the bottom of this!

build-logic/convention/src/main/java/uk/gov/onelogin/plugin/test-config.gradle.kts
elijahhezekiah-maker: Yes,  `snapshot` is not part of the TestType, but I added the string for safety measures will remove the snapshot and leave unit. 

I am still trying to understand this convenience plugin.

elijahhezekiah-maker: The Exclusion of  the `Unit  ` testType  is excluding the directory where we have the screenshot Test. 

and since we have the screenshot Test for the Features module in  the Unit directory, then test will not run . 

now wit the exclusions not including "unit" we can trigger the snapshot test using `./gradlew verifyPaparazziBuildDebug`

build-logic/convention/src/main/java/uk/gov/onelogin/plugin/test-config.gradle.kts
elijahhezekiah-maker: if we are removing the exclusion and that will make all the test run twice,  will that affect the build time when we push remotely. 

I am not sure the reason for the Plugin but it is worth finding out why we excluded them  in the first place. 

can we confirm removing the exclusion and I will push a commit with the removal of the new function and the exclusion of those testTypes

build-logic/convention/src/main/java/uk/gov/onelogin/plugin/test-config.gradle.kts
jonnyandrew: Thanks, yes if we can remove `snapshot` then there will be less chance this code causes confusion

build-logic/convention/src/main/java/uk/gov/onelogin/plugin/test-config.gradle.kts
jonnyandrew: I definitely agree that if there are tests running twice, we should work to eliminate that.

However, as the current solution/workaround is causing arguably worse problems ([DCMAW-20259](https://govukverify.atlassian.net/browse/DCMAW-20259), [DCMAW-20329](https://govukverify.atlassian.net/browse/DCMAW-20329)), I'd recommend removing it first and then addressing the problem of tests running twice.

elijahhezekiah-maker: I have removed the exclusion of testTypes in this [commit](https://github.com/govuk-one-login/mobile-android-one-login-app/pull/860/changes/011d5d63c0a3d9d3b85f1dcec417441bedf2eb47)

if you agree with the changes, I can now revert the test on the AppIntegrityErrorScreen and the PR should be ready

jonnyandrew: Thanks - it looks like the screenshot tests are still failing as expected, so yes, let's revert the change to `AppIntegrityErrorScreen`.

elijahhezekiah-maker: I have reverted  the changes on the `AppIntegrityErrorScreen`

jonnyandrew (commented):
The code changes look good, but there are failing screenshot tests. Are you planning to update those separately?

elijahhezekiah-maker:
> The code changes look good, but there are failing screenshot tests. Are you planning to update those separately?

@jonnyandrew  I wanted confirmation of the code change before recording the screenshots and I will push the commits to this PR .

features/src/test/snapshots/images/uk.gov.onelogin.features.unit.login.ui.welcome_WelcomeScreenshotTest_testScreenshot[0].png
Comment (jonnyandrew) [inline features/src/test/snapshots/images/uk.gov.onelogin.features.unit.login.ui.welcome_WelcomeScreenshotTest_testScreenshot[0].png:1]: This screenshot test seems to be running in the staging environment, whereas it was build before. Is that expected, or something that might need investigation?

What command did you use to generate the screenshot tests? I believe we'll need to use the task for the build environment if there is more than one.

Comment (elijahhezekiah-maker) [reply features/src/test/snapshots/images/uk.gov.onelogin.features.unit.login.ui.welcome_WelcomeScreenshotTest_testScreenshot[0].png:1]: I used  this command .`/gradlew :recordPaparazziStagingDebug` similar to `./gradlew recordPaparazziDebug `Bianca has showed me earlier. 

I could remove the new screenshots and then push an empty screenshot directory and see if the workflow issues still persist.

jonnyandrew (approved):

PR #874, opened 2026-06-01
feat: Enhance logs for AppInfo errors returned for /appInfo API calls

jonnyandrew:
## Changes

- Log custom error keys for /appInfo API errors

## Context

Enables better filtering in Firebase Crashlytics.

This PR supersedes #849, resolving PR comments.

DCMAW-20191

### DIFF
--- a/features/src/main/java/uk/gov/onelogin/features/appinfo/data/AppInfoRemoteSourceImpl.kt
+++ b/features/src/main/java/uk/gov/onelogin/features/appinfo/data/AppInfoRemoteSourceImpl.kt
@@ -19,6 +19,7 @@ class AppInfoRemoteSourceImpl
                         result.response as AppInfoData,
                     )
                 is ApiResponse.Offline -> AppInfoRemoteState.Offline
+                is ApiResponse.Failure -> AppInfoRemoteState.Failure("Status: ${result.status}", result.error)
                 else -> AppInfoRemoteState.Failure(APP_INFO_REMOTE_SOURCE_ERROR)
             }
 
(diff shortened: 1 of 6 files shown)


features/src/main/java/uk/gov/onelogin/features/appinfo/data/AppInfoRemoteSourceImpl.kt
```diff
@@ -19,7 +19,13 @@ class AppInfoRemoteSourceImpl
                         result.response as AppInfoData,
                     )
                 is ApiResponse.Offline -> AppInfoRemoteState.Offline
-                else -> AppInfoRemoteState.Failure(APP_INFO_REMOTE_SOURCE_ERROR)
+                is ApiResponse.Failure -> AppInfoRemoteState.Failure("Status: ${result.status}", result.error)
+                // The Loading response type is never actually returned (due to be fixed in DCMAW-20185)
+                ApiResponse.Loading ->
```
elijahhezekiah-maker, line 24: Do we want to wait until we have added the Loading state as described [here](https://govukverify.atlassian.net/browse/DCMAW-20185?search_id=434f2263-f05b-4d7f-9268-fb6ed244739a), adding this loading state and returning a failure is confusing

jonnyandrew: The `Loading` type already exists in the `ApiResponse` sealed class, it's just never returned. Before this PR, it was being included in the `else` branch and mapped to the wrong failure type.

All that I've done here is map the different response types to failure types that are more appropriate, which is in line with the goal of improving /appInfo error logging (DCMAW-20191).

In the linked ticket (DCMAW-20185) we'll remove the `Loading` state, which will eliminate this when branch, rather than make it reachable.

elijahhezekiah-maker: For future reference — once `Loading` is removed from ApiResponse [(DCMAW-20185)](https://govukverify.atlassian.net/browse/DCMAW-20185) , are we confirming that this block will simply be removed entirely? that is no `else` branch or logging needed, just the three remaining  cases `Success`, `Offline`, `Failure`?

jonnyandrew: Yes, this block will be removed entirely. The client has never returned `Offline` and `Loading` responses. In future, there will only be two response types: `Success` and `Failure`.

elijahhezekiah-maker (approved):

PR #886, opened 2026-06-12
ci: Fix build and staging app deployment workflow (develop branch)

jonnyandrew:
## Context

#819 updated the build and staging app deployment workflow to be triggered after the new test/check workflow completes.

However, the existing trigger wasn't removed. The deployment workflow is still triggered immediately on push but can't succeed as it expects the context available from the `workflow_run` trigger.

DCMAW-20463

## Changes

- Remove the `push` trigger from the build/staging app deployment workflow

### DIFF
--- a/.github/workflows/on_push_develop.yml
+++ b/.github/workflows/on_push_develop.yml
@@ -6,9 +6,6 @@ on:
     workflows: [ Run checks (tests and style checks) ]
     branches: [ develop ]
     types: [ completed ]
-  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/on_push_develop.yml
jonnyandrew: Perhaps we should rename this workflow and workflow file?

elijahhezekiah-maker (approved):

PR #887, opened 2026-06-17
refactor: DCMAW-20188 Use Hilt to instantiate StsAuthenticationProvider

jonnyandrew:
## Context

- Enables more providers similar to `AuthenticationProvider` to added to the HttpClient without bloating the `MainActivityViewModel`
- Keeps dependency injection logic in Hilt modules where it belongs

To be followed by
- #888

DCMAW-20188

### DIFF
--- a/app/src/main/java/uk/gov/onelogin/network/AuthenticationProviderModule.kt
+++ b/app/src/main/java/uk/gov/onelogin/network/AuthenticationProviderModule.kt
@@ -1,0 +1,55 @@
+package uk.gov.onelogin.network
+
+import android.content.Context
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.android.components.ViewModelComponent
+import dagger.hilt.android.qualifiers.ApplicationContext
+import uk.gov.android.network.auth.AuthenticationProvider
+import uk.gov.android.network.client.GenericHttpClient
+import uk.gov.android.onelogin.core.R
+import uk.gov.logging.api.v3.Logger
+import uk.gov.onelogin.core.navigation.domain.Navigator
+import uk.gov.onelogin.core.tokens.data.TokenRepository
+import uk.gov.onelogin.core.tokens.domain.expirychecks.IsTokenExpired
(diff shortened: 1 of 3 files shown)


app/src/main/java/uk/gov/onelogin/network/AuthenticationProviderModule.kt
```diff
@@ -0,0 +1,55 @@
...
+import uk.gov.onelogin.features.signout.domain.SignOutUseCase
+
+@Module
+@InstallIn(ViewModelComponent::class)
+object AuthenticationProviderModule {
+    @Suppress("LongParameterList")
+    @Provides
+    fun provideAuthenticationProvider(
+        activityProvider: ActivityProvider,
+        @ApplicationContext context: Context,
```
elijahhezekiah-maker, line 29: is it possible that we can access the URL in the viewmodel instead of passing the context in the function of the DI?

The natural place  for the context is in the UI layer,  we are not accessing the applicationContext to manage general app behaviour.

it is good practice to keep the DI Module lean

alex-bradbury-gds (approved):
Looks good. Think this was originally setup this way because of a cyclical dependency with the provider requiring the httpClient and the client requiring the provider... Seems like the network service setup fixes that 👍

app/src/main/java/uk/gov/onelogin/network/AuthenticationProviderModule.kt
jonnyandrew: I'm not sure how that would work, because the AuthenticationProvider sits in the data layer.

In general, `Context` may be used outside of the UI layer if it's appropriate. Lots of non-UI Android APIs need `Context`.

I agree that we should keep UI/presentation functionality out of the data layer, but in this case we're accessing a backend API URL, which isn't a UI/presentation concern.

jonnyandrew:
> this was originally setup this way because of a cyclical dependency with the provider requiring the httpClient and the client requiring the provider

I think the cyclical dependency is still there, and we're still wiring the two objects together in the `MainActivityViewModel`. This PR just pulls the provider initialisation out of the view model.

PR #941, opened 2026-07-14
fix: force update transitive bouncycastle dependency

alex-bradbury-gds:
Hoping to fix this issue: https://github.com/govuk-one-login/mobile-android-one-login-app/security/dependabot/73

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -139,3 +139,11 @@ fun setProperty(
 val composeVersion by project.extra("1.5.3")
 val intentsVersion by project.extra("3.4.0")
 val navigationVersion by project.extra("2.6.0")
+
+subprojects {
+    configurations.all {
+        resolutionStrategy {
+            force(libs.bouncycastle.bcprov.get().toString())
+        }
+    }
+}
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -141,10 +141,12 @@ val intentsVersion by project.extra("3.4.0")
 val navigationVersion by project.extra("2.6.0")
 
 subprojects {
-    dependencies {
-        constraints {
-            add("testImplementation", libs.bouncycastle.bcprov.get().toString()) {
-                because("previous versions have a CVE vulnerability (introduced transitively via robolectric)")
+    pluginManager.withPlugin("java") {
+        dependencies {
+            constraints {
+                add("testImplementation", libs.bouncycastle.bcprov.get()) {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -139,3 +139,11 @@ fun setProperty(
 val composeVersion by project.extra("1.5.3")
 val intentsVersion by project.extra("3.4.0")
 val navigationVersion by project.extra("2.6.0")
+
+subprojects {
+    configurations.all {
+        resolutionStrategy {
+            force(libs.bouncycastle.bcprov.get().toString())
```
jonnyandrew: This will force the exact version of bouncycastle, meaning that it won't be able to upgraded transitively, which I don't think is what we'd want.

Gradle [recommends](https://docs.gradle.org/current/userguide/resolution_rules.html#sec:forcing-dep-versions) using [dependency constraints](https://docs.gradle.org/current/userguide/dependency_constraints.html#sec:adding-constraints-transitive-deps) for this as a less brute-force approach - please could we try that?

build.gradle.kts
```diff
@@ -139,3 +139,11 @@ fun setProperty(
 val composeVersion by project.extra("1.5.3")
 val intentsVersion by project.extra("3.4.0")
 val navigationVersion by project.extra("2.6.0")
+
+subprojects {
```
jonnyandrew, line 143: Ideally we can avoid using `subprojects` for performance reasons (see https://docs.gradle.org/current/userguide/configuration_on_demand.html#sec:decoupled_projects)

Instead of applying this in the root build.gradle.kts, can we apply it through a plugin, maybe `uk.gov.pipelines.android-lib-config` and `uk.gov.pipelines.android-lib-config`?

That would also let all our projects inherit the same secure dependency constraints.

alex-bradbury-gds: yep, it's a good plan, we will refactor to this approaching assuming this fix 'works'

build.gradle.kts
alex-bradbury-gds: done 👍

jonnyandrew (approved):

build.gradle.kts
jonnyandrew: Sounds good - and by the way, I've raised a PR with what I'm proposing in case it's helpful:
- https://github.com/govuk-one-login/mobile-android-one-login-app/pull/942

PR #871, opened 2026-05-28
feat: prompt re auth when wallet store id is missing from device storage

elijahhezekiah-maker:
feat:  re-auth when wallet store id is missing

- added GetWalletStoreIdImpl.kt usecase
- updated RefreshExchangeImpl and HandleLocalLoginImpl to validate wallet store ID
- added ValidateWalletStoreID returning true if  wallet store ID is present
- added test for SplashScreenViewModelTest.kt
- added  test for GetWalletStoreIdImplTest.kt
- added test for RefreshExchangeImplTest.kt
- added RemoveFromOpenSecureStore 
- added RemoveFromOpenSecureStoreImpl

**Evidence**
**Acceptance Criteria**

AC 2

GIVEN a returning user opens the One Login app via the SplashScreen
AND the Wallet Store ID does not exist in shared prefs
AND the user is online
WHEN the check is performed
THEN the user is prompted to re-authenticate

https://github.com/user-attachments/assets/78625e1f-8989-4a32-b692-fdff123d1ed2

AC 3 

GIVEN a returning user opens the One Login app via the SplashScreen
AND the Wallet Store ID does exist in shared prefs
AND the user is online
WHEN the check is performed
THEN the user continues through the existing app flow without interruption

https://github.com/user-attachments/assets/927c536e-0262-4144-9fa5-e5f0ec2a45c6

AC 4

GIVEN a returning user opens the One Login app via the SplashScreen
AND the Wallet Store ID does not exist in shared prefs
AND the user is offline
WHEN the check is performed
THEN the user is directed to the ‘Network connection error’ page

https://github.com/user-attachments/assets/ad343f44-d13e-4709-8760-9cbed6dae94f

AC 5 

GIVEN a returning user opens the One Login app via the SplashScreen
AND the Wallet Store ID does exist in shared prefs
AND the user is offline
WHEN the check is performed
THEN the user continues through the app as normal (and can access the offline wallet)

https://github.com/user-attachments/assets/8da6f5c2-ac99-4b3b-84f9-deb56d7b5ee3

Resolves: [DCMAW-20192](https://govukverify.atlassian.net/browse/DCMAW-20192)

[DCMAW-20192]: https://govukverify.atlassian.net/browse/DCMAW-20192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #946, opened 2026-07-16
feat: DCMAW-20393 Supply wallet store ID to Wallet SDK

elijahhezekiah-maker:
feat: Supply wallet store ID to Wallet SDK

Introduces support for passing a walletStoreId to the Wallet SDK as part of the Encrypting
  Wallet Credentials feature.
  
  - Introduce WalletScreenViewModel to retrieve and set wallet store ID
  - Call setWalletStoreId on lifecycle RESUMED via WalletScreen
  - Log error with structured keys if retrieval fails

As part of the Encrypting Wallet Credentials feature, Wallet requires the new walletStoreID.  This change allows the One Login app to retrieve and supply that ID 

Resolves: DCMAW-20393


features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreen.kt
```diff
@@ -1,16 +1,27 @@
...
     deepLinkRoute: Boolean,
     setDisplayContentAsFullScreen: (Boolean) -> Unit,
     viewModel: WalletScreenViewModel = hiltViewModel(),
 ) {
+    val lifecycleOwner = LocalLifecycleOwner.current
+    LaunchedEffect(lifecycleOwner) {
+        lifecycleOwner.repeatOnLifecycle(Lifecycle.State.RESUMED) {
+            viewModel.setWalletStoreId()
+        }
+    }
```
jonnyandrew: This is business logic that the UI shouldn't need to be involved with. Please can we remove this?

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenViewModel.kt
```diff
@@ -1,13 +1,52 @@
...
     constructor(
         val walletSdk: WalletSdk,
-    ) : ViewModel()
+        private val getWalletStoreId: GetWalletStoreId,
+        private val logger: Logger
+    ) : ViewModel(),
+        LogTagProvider {
+        fun setWalletStoreId() {
+            viewModelScope.launch {
+                runCatching { getWalletStoreId() }
```
jonnyandrew: `getWalletStoreId()` doesn't throw, so `runCatching` is redundant.

Please can we remove the try/catch and error logging?

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenViewModel.kt
```diff
@@ -1,13 +1,52 @@
...
+                    .onSuccess { storeId -> storeId?.let {
+                        walletSdk.setWalletStoreId(it) } }
```
jonnyandrew: `setWalletStoreId()` must be called before `WalletApp()` but I can't see that we are guaranteeing this order here.

Please can we ensure this constraint is guaranteed by our implementation, and that the required order clear in the code with comments and tests (so that it can't break in future)?

The way that I'd suggest implementing this is:

#### Expose `state` Flow that keeps the `WalletApp()` unavailable until `WalletSdk` is fully initialised
```kt
    private val _state: MutableStateFlow<WalletScreenState> = MutableStateFlow(
        WalletScreenState.Loading
    )
    private val state: Flow<WalletScreenState> = _state.asFlow()
```

#### Initialise the `WalletSdk` in `ViewModel.init()`
```kt
    init {
        viewModelScope.launch {
            val walletAppDisplayer = prepareWalletSdkForDisplay()
           
            _state.value = WalletScreenState.Display(walletAppDisplayer)
        }
    }

    private suspend fun prepareWalletSdkForDisplay(): WalletAppDisplayer {
        val walletStoreId = getWalletStoreId()

        requireNotNull(walletStoreId) {
            "Wallet Store ID must not be null. This is guaranteed by ValidateWalletStoreId"
        }

        // Must be called before WalletSdk.WalletApp()
        walletSdk.setWalletStoreId(walletStoreId)

        return WalletAppDisplayer(walletSdk)
    }
}
```

#### Avoid giving the UI direct access to the `WalletSdk`
```kt
interface WalletAppDisplayer {
    @Composable
    fun WalletApp()
}
```

features/src/test/java/uk/gov/onelogin/features/unit/wallet/WalletScreenViewModelTest.kt
```diff
@@ -0,0 +1,84 @@
...
+import uk.gov.logging.api.v3.matchers.LogEntryMatchers.hasMessage
+import uk.gov.logging.api.v3.matchers.MemorisedLoggerMatchers.hasSize
+import uk.gov.onelogin.core.tokens.domain.retrieve.GetWalletStoreId
+import uk.gov.onelogin.features.extensions.CoroutinesTestExtension
+import uk.gov.onelogin.features.extensions.InstantExecutorExtension
+import uk.gov.onelogin.features.wallet.ui.WalletScreenViewModel
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@ExtendWith(InstantExecutorExtension::class, CoroutinesTestExtension::class)
+class WalletScreenViewModelTest {
```
jonnyandrew, line 23: I haven't reviewed the tests as they would need to change with the implementation

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenViewModel.kt
elijahhezekiah-maker: we cannot avoid giving the UI direct  access to the wallet sdk  since  the deeplink route needs to be set
directly on `WalletScreen` since users can add credentials via deeplink, which is why the boolean is
passed through the `walletScreen's `constructor.

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenViewModel.kt
elijahhezekiah-maker: Considering the possibility of tracking any error or exception as
  we begin to extend `WalletScreenViewModel `with more features, the
  `runCatching` has been added to allow for more observability within
  this component.
  
Do we want to reconsider, open to suggestions and confirmation

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreen.kt
elijahhezekiah-maker: yes it is worth considering  rewriting this implementation we could improve the abstraction here by introducing
  an `onResumed()` function on the ViewModel rather than calling
 ` setWalletStoreId()` directly from the UI:
  
```

lifecycleOwner.repeatOnLifecycle(Lifecycle.State.RESUMED) {
      viewModel.onResumed()
  }
  
  fun onResumed() {
      viewModelScope.launch {
          prepareWalletSdkForDisplay()
      }
  }
  
```
   This ensures the UI layer does not have the business logic
  
The reason we're triggering this on `RESUMED` rather than in the
  `init` block is to correctly handle the re-auth flow. If a user
  doesn't have a `walletStoreId`, re-auth is triggered. 
  
  Since the ViewModel instance may already exist when the user returns to
  `WalletScreen`, the `init` block won't run again,meaning
  `setWalletStoreId` would never be called with the updated value.
  Tying the call to the `RESUMED` lifecycle event ensures it's always
  invoked with the latest store ID on each screen visit.

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenViewModel.kt
jonnyandrew: It's worth adding try/catch when the function we're calling will throw an exception, and in that case yes, logging the error would give observability.

But in this case the function we're calling doesn't throw, making a lot of this code redundant as it will never be called.

Please can you take a look at this thread where we've discussed similar things before, and let me know if that makes sense in this context? https://gds.slack.com/archives/C07K1GFTGP3/p1781780131882539?thread_ts=1779981259.830949&cid=C07K1GFTGP3

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenViewModel.kt
jonnyandrew: I'm not sure I understand the relevance of the deep link route here. Perhaps we can discuss the final point about UI talking directly to `WalletSdk` separately, after we've addressed the order of logic.

Does my main point about order of `setWalletStoreId()` and `WalletSdk.WalletApp()` make sense?

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreen.kt
jonnyandrew: Ah OK, I think that's a misunderstanding then - there's no need to repeat `setWalletStoreId()` each time the app is resumed because the value of the Wallet Store ID won't change.

We just need to make sure we've initialised the instance of `WalletSdk` that the view model has, before displaying `WalletApp()` for the first time.

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenViewModel.kt
elijahhezekiah-maker: > I'm not sure I understand the relevance of the deep link route here. 

This is in relation to your comment about;

> Avoid giving the UI direct access to the WalletSdk

> Does my main point about order of setWalletStoreId() and WalletSdk.WalletApp() make sense?

your point about  setWalletStoreId() and WalletSdk.WalletApp() is gradually becoming clear, the only thing I need clarification on  is the code snippet provided. 

we have a WalletScreenState added to  `walletScreenViewmodel`. 
```
 private val _state: MutableStateFlow<WalletScreenState> = MutableStateFlow(
        WalletScreenState.Loading
    )
```

are we  creating  a sealed class to handle the different state before the WalletSdk initialisation then

jonnyandrew: Yes, as we can no longer call `WalletSdk.WalletApp()` immediately, so I'm proposing that we introduce a loading state for the screen:

```kt
sealed interface WalletScreenState {
    data object Loading: WalletScreenState

    data class Display(
        val walletAppDisplayer: WalletAppDisplayer,
    ) : WalletScreenState
}
```

elijahhezekiah-maker: Now that we have a `Loading state` what will be the expected UI behaviour?

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenViewModel.kt
```diff
@@ -1,13 +1,60 @@
...
+
+    /**
+     * Retrieves the wallet store ID and sets it on the SDK before returning the Wallet Sdk for Initialisation.
+     *
+     */
+    private suspend fun prepareWalletSdkForDisplay(): WalletAppDisplayer {
+        val walletStoreId = getWalletStoreId()
+
+        // Must be called before WalletSdk.WalletApp()
+        walletStoreId?.let { walletSdk.setWalletStoreId(it) }
```
jonnyandrew: This null-safe access hides an unrecoverable error situation that should not be possible unless there has been a programming error.

If the wallet store ID is null, then it means the wallet store ID validation (#871) has failed and the user has not been logged out as they should have. In that case its probably safest to crash so that the wallet store ID validation has another chance to run.

Please can we use `requireNotNull` here (see https://github.com/govuk-one-login/mobile-android-one-login-app/pull/946#discussion_r3594357235)?

```kt
requireNotNull(walletStoreId) {
    "Wallet Store ID must not be null. This is guaranteed by ValidateWalletStoreId"
}
```

Please can we also cover this scenario with a test?

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenViewModel.kt
```diff
@@ -1,13 +1,60 @@
...
+    init {
+        viewModelScope.launch(dispatcher) {
+            _state.value = WalletScreenState.Display(prepareWalletSdkForDisplay())
+        }
+    }
+
+    /**
+     * Retrieves the wallet store ID and sets it on the SDK before returning the Wallet Sdk for Initialisation.
+     *
+     */
```
jonnyandrew, line 58: This comment isn't adding much value as it states exactly what happens in the function.

A comment here could still be useful but please could we explain why this code is needed?

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenViewModel.kt
```diff
@@ -40,7 +37,7 @@ class WalletScreenViewModel
     val state: StateFlow<WalletScreenState> = _state.asStateFlow()
 
     init {
-        viewModelScope.launch(dispatcher) {
```
jonnyandrew: @elijahhezekiah-maker FYI, this would have been redundant because the main scheduler is hard coded in the ViewModel scope.

I have pushed a commit to refactor the tests to remove this: https://github.com/govuk-one-login/mobile-android-one-login-app/pull/946/commits/1c76a72215bd32df568c8f6f22c9d567fd0529fb

Please can you take a look at this commit and also https://developer.android.com/kotlin/coroutines/test#setting-main-dispatcher

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreenState.kt
```diff
@@ -0,0 +1,23 @@
+package uk.gov.onelogin.features.wallet.ui
+
+import uk.gov.onelogin.core.ui.wallet.WalletAppDisplayer
+
+/**
+ * The sealed class is  the UI state of [WalletScreen].
+ *
+ *
+ * - [Loading] is the initial state. It acts as a structural guard to ensure the Wallet SDK
+ *   is never rendered before [WalletAppDisplayer]
```
jonnyandrew, line 10: This sentence doesn't seem to be complete

jonnyandrew (dismissed):

app/src/main/java/uk/gov/onelogin/wallet/WalletModule.kt
```diff
@@ -23,29 +23,29 @@ import uk.gov.onelogin.core.ui.wallet.WalletAppDisplayerImpl
...
-                )
-            return WalletSdkImpl(navigator, config, context)
-        }
+    @Suppress("LongParameterList")
+    @Provides
+    fun provideWalletSdk(
+        @ApplicationContext
+        context: Context,
+        navigator: Navigator,
+        defaultNetworkService: DefaultNetworkServiceV1,
```
jonnyandrew: Please can we inject `NetworkService` (V1) (interface) here, rather than the `DefaultNetworkService` implementation?

jonnyandrew (approved):

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreen.kt
```diff
@@ -1,18 +1,28 @@
...
     setDisplayContentAsFullScreen: (Boolean) -> Unit,
     viewModel: WalletScreenViewModel = hiltViewModel(),
 ) {
+
+
     key(deepLinkRoute) {
-        viewModel.walletSdk.WalletApp()
+        when (val state = viewModel.state.collectAsState().value) {
+            is WalletScreenState.Loading -> {
+                LoadingScreen { }
```
jonnyandrew, line 21: Just a reminder that we discussed that the loading UI shouldn't flicker - have you tested this manually?

elijahhezekiah-maker: Yes,  the loading UI is not flickering when adding credentials to the wallet.

PR #954, opened 2026-07-22
ci: DCMAW-21639 Wait for release approval before generating version code

jonnyandrew:
## Context

Usually, the jobs in the `on_push_release` workflow wait for a while before being deployed. However, the versioning job runs immediately upon raising the PR. As the version code is based on the current timestamp, it can go stale as soon as a new build is published by a different branch.

This PR intends to reduce the risk of that happening by delaying the version code generation, gating it behind the same approval needed for deployment.


.github/workflows/on_push_release.yml
```diff
@@ -14,6 +14,10 @@ jobs:
   set-up:
     name: Set-up and generate version name and version code
     runs-on: ubuntu-22.04
+    # This job doesn't need the environment.
+    # It is behind an approval step so that the version code isn't generated immediately and
+    # is less likely to go stale.
+    environment: release
```
jonnyandrew, line 20: There is no way to apply `environment` to the whole workflow (https://github.com/orgs/community/discussions/174381), meaning the approver would need to approve both jobs now.

The `environment` grants access to any variables that are part of the `release` environment to this job.

I'm unsure whether there are any variables in the `release` environment? If there are not, then we could remove the `environment: release` from the deployment jobs and use this job as the gate (as mentioned in https://github.com/orgs/community/discussions/174381.

Other options:
- Move the version generation into the deployment job (and sacrifice consistency between the Build/Staging versions due to timing)
- Accept that the version code may go stale and we'll need to rerun the job before approving the workflow

elijahhezekiah-maker (approved):

PR #955, opened 2026-07-23
refactor: Migrate to NetworkService V2 in StsAuthenticationProvider

jonnyandrew:
DCMAW-21564

### DIFF
--- a/features/src/main/java/uk/gov/onelogin/features/network/provider/StsAuthenticationProvider.kt
+++ b/features/src/main/java/uk/gov/onelogin/features/network/provider/StsAuthenticationProvider.kt
@@ -2,19 +2,21 @@ package uk.gov.onelogin.features.network.provider
 
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
-import kotlinx.serialization.json.Json
-import uk.gov.android.network.api.ApiRequest
-import uk.gov.android.network.api.ApiResponse
 import uk.gov.android.network.api.ApiResponseException
+import uk.gov.android.network.api.v2.ApiRequest
+import uk.gov.android.network.api.v3.ApiResponse
 import uk.gov.android.network.auth.AuthenticationProvider
 import uk.gov.android.network.auth.AuthenticationResponse
 import uk.gov.android.network.client.GenericHttpClient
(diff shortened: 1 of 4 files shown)


features/src/test/java/uk/gov/onelogin/features/unit/network/provider/StsAuthenticationProviderTest.kt
```diff
@@ -60,7 +64,7 @@ class StsAuthenticationProviderTest {
         whenever(mockTokenRepository.getTokenResponse()).thenReturn(loginTokens)
         whenever(mockActivityProvider.getCurrentActivity()).thenReturn(mockFragmentActivity)
         wheneverBlocking { mockIsAccessTokenExpired.invoke() }.thenReturn(false)
-        wheneverBlocking { mockHttpClient.makeRequest(any()) }.thenReturn(ApiResponse.Loading)
```
jonnyandrew, line 63: Previously, the default response for `mockHttpClient` in this test was `ApiResponse.Loading`.

`ApiResponse.Loading` is treated as an error in most cases because the `GenericHttpClient` V1 implementation never emits it.

The following tests depended on this causing a failure:

- `access token expired, refresh exchange has failed with offline network`
- `access token expired, refresh exchange has failed with user cancelled bio prompt`

I've changed the setup in these tests so that the `httpClient` returns 400.

features/src/main/java/uk/gov/onelogin/features/network/provider/StsAuthenticationProvider.kt
```diff
@@ -163,51 +177,20 @@ class StsAuthenticationProvider(
...
-                    logger.error(
-                        loginException::class.java.simpleName,
-                        e.message.toString(),
-                        loginException
-                    )
-                    AuthenticationResponse.Failure(e)
-                }
-
-            // Check response for account intervention
-            is ApiResponse.Failure -> {
```
jonnyandrew, line 193: I was slightly unsure where to move this comment. Is it referring to the 'invalid grant' / 400 / re-auth case below? Or API failure in general?

elijahhezekiah-maker (approved):

PR #968, opened 2026-07-31
refactor: Migrate to NetworkService V2 for refresh exchange DPoP generation

jonnyandrew:
DCMAW-21564

### DIFF
--- a/features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
+++ b/features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
@@ -5,10 +5,7 @@ import androidx.fragment.app.FragmentActivity
 import dagger.hilt.android.qualifiers.ApplicationContext
 import io.ktor.client.request.forms.FormDataContent
 import io.ktor.http.Parameters
-import kotlinx.serialization.json.Json
 import uk.gov.android.authentication.integrity.pop.SignedPoP
-import uk.gov.android.authentication.login.refresh.DemonstratingProofOfPossessionManager
-import uk.gov.android.authentication.login.refresh.SignedDPoP
 import uk.gov.android.network.api.v2.ApiRequest
 import uk.gov.android.network.api.v3.ApiResponse
 import uk.gov.android.network.service.NetworkingException
@@ -49,7 +46,6 @@ class RefreshExchangeImpl
         private val isRefreshTokenExpired: IsTokenExpired,
         private val networkService: NetworkService,
         private val appIntegrity: AppIntegrity,
-        private val dPoPManager: DemonstratingProofOfPossessionManager,
         private val getFromEncryptedSecureStore: GetFromEncryptedSecureStore,
         private val saveTokenExpiry: SaveTokenExpiry,
         private val tokenRepository: TokenRepository,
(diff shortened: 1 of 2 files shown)


features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
```diff
@@ -221,50 +217,35 @@ class RefreshExchangeImpl
                     R.string.stsUrl,
                     context.getString(R.string.tokenExchangeEndpoint),
                 )
-            // Attempt to generate DPoP
-            val dPoP = dPoPManager.generateDPoP(authUrl)
```
elijahhezekiah-maker, line 225: The refresh token request already uses withRefreshDPoP = true in the network service for proof of possession. Introducing appIntegrity.getProofOfPossession() (Firebase PoP) alongside that creates
two different proof mechanisms in the same class, which is confusing.

Could we please keep this consistent and rely solely on DPoP for the proof of possession here? If the Firebase PoP is genuinely required in addition to DPoP, could we please document clearly in the
class the  distinct role each plays

jonnyandrew: > Introducing appIntegrity.getProofOfPossession() (Firebase PoP) alongside that creates two different proof mechanisms in the same class, which is confusing.
> Could we please keep this consistent and rely solely on DPoP for the proof of possession here?

We are using `appIntegrity.getProofOfPossession()` to generate the proof of possession for the client attestation. But that is already in place - we aren't introducing anything new in this PR.

The API call needs all of client attestation, client attestation PoP and DPoP.

> Could we please document clearly in the class the distinct role each plays

As the Networking library is now responsible for these things, would you agree that we can maintain the documentation there centrally?

Please can you take a look at the following classes KDoc and see what you think of what we already have? I think there's room for improvement in the `RequestConfigBuilder` docs, but let me know if you see any other gaps and I'll try to address them.

- [ClientAttestationProvider](https://github.com/govuk-one-login/mobile-android-networking/blob/56524ed76dc9080b5ee8a2ec942d7f9484e0d74b/network/src/main/java/uk/gov/android/network/attestation/ClientAttestationProvider.kt)
- [DPoPProvider](https://github.com/govuk-one-login/mobile-android-networking/blob/56524ed76dc9080b5ee8a2ec942d7f9484e0d74b/network/src/main/java/uk/gov/android/network/dpop/DPoPProvider.kt)
- [RequestConfigBuilder](https://github.com/govuk-one-login/mobile-android-networking/blob/56524ed76dc9080b5ee8a2ec942d7f9484e0d74b/network/src/main/java/uk/gov/android/network/client/config/RequestConfigBuilder.kt)

elijahhezekiah-maker (approved):

features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
jonnyandrew: Clarified in https://gds.slack.com/archives/C07K1GFTGP3/p1785839371032769?thread_ts=1785496524.546849&cid=C07K1GFTGP3

PR #969, opened 2026-07-31
refactor: Migrate to NetworkService V2 for refresh exchange client attestation

jonnyandrew:
DCMAW-21564

### DIFF
--- a/features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
+++ b/features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
@@ -51,7 +51,6 @@ class RefreshExchangeImpl
         private val timeProvider: TimeProvider,
         private val validateWalletStoreId: ValidateWalletStoreId,
     ) : RefreshExchange {
-        // Initialise token and client attestation field
         private var refreshToken = ""
         private var idToken = ""
         private var areChecksSuccessful = false
@@ -64,7 +63,6 @@ class RefreshExchangeImpl
             if (!getPersistentId().isNullOrEmpty()) {
                 // Check Refresh token is NOT expired and wallet store ID is valid
                 if (!isRefreshTokenExpired() && validateWalletStoreId()) {
-                    // Attempt to get Client Attestation
                     getTokensFromSecureStore(context, handleResult)
                 } else {
                     // When Refresh token is invalid or wallet store ID is missing
@@ -240,8 +238,6 @@ class RefreshExchangeImpl
 
         companion object {
             const val REFRESH_ERROR_TAG = "Refresh Exchange Tokens Error"
-            const val EMPTY_MSG = "Refresh Token Exchange Error - Error message was null."
-            const val ATTESTATION_POP_GENERATE_ERROR = "Couldn't generate App Integrity PoP."
             private const val CONTENT_TYPE_LABEL = "Content-Type"
             private const val CONTENT_TYPE_VALUE = "application/x-www-form-urlencoded"
             private const val GRANT_TYPE_LABEL = "grant_type"
--- a/features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
+++ b/features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
@@ -80,11 +80,15 @@ class RefreshExchangeImpl
             if (areChecksSuccessful) handleResult(makeRefreshTokenCall())
         }
 
+        /**
+         * Get the saved refresh token and ID token and set [refreshToken] and [idToken].
+         *
+         * If successful, it sets [areChecksSuccessful] to `true`, otherwise it calls [onFailure].
+         */
         private suspend fun getTokensFromSecureStore(
             context: FragmentActivity,
             onFailure: (RefreshExchangeResult) -> Unit,
         ) {
-            // Attempt to retrieve the Refresh token from the secure store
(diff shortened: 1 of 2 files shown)


features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
```diff
@@ -209,52 +153,30 @@ class RefreshExchangeImpl
...
-                        pop.reason,
-                        pop.error ?: fallbackExp,
-                    )
-                    // Throw error if Client Attestation PoP couldn't be generated
-                    throw fallbackExp
-                }
+            return networkService.makeRequest<RefreshExchangeApiResponse>(
+                apiRequest = request,
+            ) {
+                withAttestation = true
```
jonnyandrew, line 178: Adding `withAttestation = true` is how we make `networkService` get the client attestation for us

features/src/test/java/uk/gov/onelogin/features/unit/login/domain/refresh/RefreshExchangeImplTest.kt
```diff
@@ -249,30 +249,16 @@ class RefreshExchangeImplTest {
...
             whenever(appIntegrity.getProofOfPossession())
                 .thenReturn(SignedPoP.Failure("Failure"))
 
             val result = getTokens()
 
-            assertErrorLogged(ATTESTATION_POP_GENERATE_ERROR)
+            assertErrorLogged("Attestation provider failed to fetch client attestation", ClientAttestationException::class.java)
             assertNothingSaved()
-            assertEquals(RefreshExchangeResult.ReauthRequired, result)
+            assertEquals(RefreshExchangeResult.ClientAttestationFailure, result)
```
jonnyandrew, line 263: This behaviour has changed:

| Scenario | Before | After |
|-|--------|-------|
| Fail to get the client attestation itself | Go to app integrity error | Go to app integrity error |
| Fail to generate the client attestation proof-of-possesion | Go to re-auth | **Go to app integrity error** |

We need to confirm what the behaviour should be before going ahead with this change.

We'd need more information from the `NetworkService` error to maintain the current behaviour.

Further discussion: https://govukverify.atlassian.net/browse/DCMAW-21564?focusedCommentId=333519

features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
elijahhezekiah-maker: Can we consider removing all references to Client Attestation from this class, since it's now handled in the network layer? I think it would be good encapsulation.

features/src/test/java/uk/gov/onelogin/features/unit/login/domain/refresh/RefreshExchangeImplTest.kt
elijahhezekiah-maker: From the perspective of this class, a client attestation failure is a network failure, and the expected result inline with app behaviour ,should be ReauthRequired. 

These tests are asserting indirect attestation concerns that belong in network layer tests, not here. 

Can we consider reverting these and ensuring the tests reflect the actual responsibilities of RefreshExchangeImpl?

elijahhezekiah-maker (approved):

features/src/test/java/uk/gov/onelogin/features/unit/login/domain/refresh/RefreshExchangeImplTest.kt
jonnyandrew: I discussed with product: https://govukverify.atlassian.net/browse/DCMAW-21564?focusedCommentId=335477

We discovered that on iOS, when the app fails to generate client attestation PoP it will send the user to the app integrity error. We'll align with that behaviour.

features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
jonnyandrew: I've removed a few stray comments that referenced client attestation: https://github.com/govuk-one-login/mobile-android-one-login-app/pull/969/commits/f28368496102351cdd9a4618dc6a8ae739dae1e2

Is that what you meant?

features/src/test/java/uk/gov/onelogin/features/unit/login/domain/refresh/RefreshExchangeImplTest.kt
elijahhezekiah-maker: Expected implementation to handle new behaviour as confirmed in this [conversation](https://gds.slack.com/archives/C07K1GFTGP3/p1786547330379989).

elijahhezekiah-maker (approved):

features/src/main/java/uk/gov/onelogin/features/login/domain/refresh/RefreshExchangeImpl.kt
elijahhezekiah-maker: My comment is now redundant considering, the new direction we are going of aligning with IOS behaviour,

could we add some Kdoc on the `getTokensFromSecureStore` function

jonnyandrew: Thanks, I've done that: https://github.com/govuk-one-login/mobile-android-one-login-app/pull/969/commits/44001234c285757137e29cdd72993a1a691297ed

PR #977, opened 2026-08-03
refactor: Remove circular dependency between authentication provider and network service

jonnyandrew:
DCMAW-21564

### DIFF
--- a/app/src/main/java/uk/gov/onelogin/network/NetworkModule.kt
+++ b/app/src/main/java/uk/gov/onelogin/network/NetworkModule.kt
@@ -9,8 +9,11 @@ import dagger.Provides
 import dagger.hilt.InstallIn
 import dagger.hilt.android.qualifiers.ApplicationContext
 import dagger.hilt.components.SingletonComponent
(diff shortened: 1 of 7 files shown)


app/src/main/java/uk/gov/onelogin/network/NetworkModule.kt
```diff
@@ -54,9 +57,16 @@ interface NetworkModule {
...
+            authenticationProvider: AuthenticationProvider,
+            dPoPProvider: DPoPProvider,
+            clientAttestationProvider: ClientAttestationProvider,
         ): DefaultNetworkService =
-            DefaultNetworkService(ktorHttpClient)
+            DefaultNetworkService(ktorHttpClient).apply {
+                setAuthenticationProvider(authenticationProvider)
+                setClientAttestationProvider(clientAttestationProvider)
+                setDPoPProvider(dPoPProvider)
+            }
```
jonnyandrew, line 69: This PR improves things by ensuring that `NetworkService` is immediately usable. 

The trade-off is that `StsAuthenticationProvider` needs to be more complex, as it now uses the lower level `GenericHttpClient` (v2).

PR #986, opened 2026-08-05
build: Improve noisy error logs when gradle properties are missing

jonnyandrew:
## Context

Makes it easier to find errors in CI build logs.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -48,12 +48,11 @@ fun fetchGithubCredentials(): Pair<String, String> {
 
     return try {
         gprUser.get() to gprToken.get()
-    } catch (exception: MissingValueException) {
+    } catch (_: MissingValueException) {
         logger.warn(
             "Could not find 'Github Package Registry' properties. Refer to the proceeding " +
                 "location for instructions:\n\n" +
                 "${rootDir.path}/docs/developerSetup/github-authentication.md\n",
-            exception
         )
 
         System.getenv("USERNAME") to System.getenv("TOKEN")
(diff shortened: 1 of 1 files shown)


settings.gradle.kts
```diff
@@ -48,12 +48,11 @@ fun fetchGithubCredentials(): Pair<String, String> {
 
     return try {
         gprUser.get() to gprToken.get()
-    } catch (exception: MissingValueException) {
+    } catch (_: MissingValueException) {
         logger.warn(
             "Could not find 'Github Package Registry' properties. Refer to the proceeding " +
                 "location for instructions:\n\n" +
                 "${rootDir.path}/docs/developerSetup/github-authentication.md\n",
-            exception
```
jonnyandrew, line 56: Removes the stack trace from the log because it's noisy and not very helpful

elijahhezekiah-maker: can we keep this, as sometimes we do have missing properties issues that affects builds in the pipeline

jonnyandrew: The reason I'm suggesting to remove the exception stack trace is that it won't give us any extra information beyond:

```
org.gradle.api.internal.provider.MissingValueException: Cannot query the value of Gradle property 'gpr.user' because it has no value available.
```

Then there are around 400 lines of logs that we'll never look at.

Would it work for you if I change this so that the name of the missing property (`gpr.user` or `gpr.token`) is logged in the message?

elijahhezekiah-maker: Yes that will work, if we could have the missing property logged that will be helpful,

jonnyandrew: Ok, I've made that change https://github.com/govuk-one-login/mobile-android-one-login-app/pull/986/commits/5c1f6578b3ee38b74f89490bd0845fd30ac9125a

elijahhezekiah-maker (approved):

PR #1006, opened 2026-08-18
ci: DCMAW-21639 Delay version code generation until deployment

jonnyandrew:
## Context

Usually, the jobs in the `on_push_release` workflow wait for a while before being deployed. However, the versioning job runs immediately upon raising the PR. As the version code is based on the current timestamp, it can go stale as soon as a new build is published by a different branch.

This PR intends to reduce the risk of that happening by delaying the version code generation, moving it into the same job that deploys the app builds.

### DIFF
--- a/.github/workflows/on_push_release.yml
+++ b/.github/workflows/on_push_release.yml
@@ -12,10 +12,9 @@ concurrency:
 
 jobs:
   set-up:
-    name: Set-up and generate version name and version code
+    name: Set-up and generate version name
     runs-on: ubuntu-24.04
     outputs:
-      version-code: ${{ steps.version-code.outputs.version-code }}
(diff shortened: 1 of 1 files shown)


.github/workflows/on_push_release.yml
```diff
@@ -94,6 +89,10 @@ jobs:
         with:
           actions-role-arn: ${{ secrets.GITHUBRUNNER_EC2_ACTIONS_ROLE_ARN }}
 
+      - name: Generate version code
+        id: version-code
+        uses: ./mobile-android-pipelines/actions/generate-version-code
```
jonnyandrew, line 94: After this change, the version code between Build and Staging builds may be different.

jonnyandrew:
Closing in favour of https://github.com/govuk-one-login/mobile-android-one-login-app/pull/954

PR #1007, opened 2026-08-18
fix: DCMAW-22138 Workaround for crash occurring when wallet store ID initialisation is cancelled

jonnyandrew:
## Context

Fixes a crash when attempting to add a wallet credential from a deep link from a cold start (DCMAW-22138).

The root cause, that Secure Store consumes coroutine cancellations and doesn't propagate them to calling code (DCMAW-22143), can be fixed separately.

## Evidence

[Screen_recording_20260818_163650.webm](https://github.com/user-attachments/assets/c15c1bca-afcb-4258-a8a3-679d897e414c)

### DIFF
--- a/features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreen.kt
+++ b/features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreen.kt
@@ -3,22 +3,24 @@ package uk.gov.onelogin.features.wallet.ui
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.key
-import androidx.hilt.navigation.compose.hiltViewModel
+import androidx.hilt.lifecycle.viewmodel.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import uk.gov.onelogin.core.ui.pages.loading.LoadingScreen
+import uk.gov.onelogin.core.ui.pages.loading.LoadingScreenAnalyticsViewModel
 
 @Composable
 fun WalletScreen(
     deepLinkRoute: Boolean,
     setDisplayContentAsFullScreen: (Boolean) -> Unit,
     viewModel: WalletScreenViewModel = hiltViewModel(),
+    loadingAnalyticsViewModel: LoadingScreenAnalyticsViewModel = hiltViewModel(),
 ) {
 
 
     key(deepLinkRoute) {
         when (val state = viewModel.state.collectAsState().value) {
             is WalletScreenState.Loading -> {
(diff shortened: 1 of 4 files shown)


features/src/test/java/uk/gov/onelogin/features/unit/wallet/WalletScreenViewModelTest.kt
```diff
@@ -100,6 +106,48 @@ class WalletScreenViewModelTest {
...
+                    getWalletStoreId = getWalletStoreIdCancelling,
+                    walletAppDisplayer = walletAppDisplayer,
+                )
+            }
+        }
+    }
+
+    // This test is redundant once DCMAW-22143 is fixed.
+    @Test
+    fun `given wallet store ID is null because coroutine cancellation isn't propogated, it does not crash`() {
```
jonnyandrew, line 129: This test fails without the workaround

features/src/test/java/uk/gov/onelogin/features/unit/wallet/WalletScreenViewModelTest.kt
```diff
@@ -100,6 +106,48 @@ class WalletScreenViewModelTest {
         )
     }
 
+    @Test
+    fun `given wallet store ID is null but coroutines are cancelled before retrieval, it does not crash`() {
```
jonnyandrew, line 110: This test reproduces the scenario where secure store _does_ propagate a cancellation exception to us. This test passes even without the workaround.

features/src/main/java/uk/gov/onelogin/features/wallet/ui/WalletScreen.kt
```diff
@@ -3,22 +3,24 @@ package uk.gov.onelogin.features.wallet.ui
...
     viewModel: WalletScreenViewModel = hiltViewModel(),
+    loadingAnalyticsViewModel: LoadingScreenAnalyticsViewModel = hiltViewModel(),
 ) {
 
 
     key(deepLinkRoute) {
         when (val state = viewModel.state.collectAsState().value) {
             is WalletScreenState.Loading -> {
-                LoadingScreen { }
+                LoadingScreen(analyticsViewModel = loadingAnalyticsViewModel) { }
```
jonnyandrew, line 23: This change is needed because a side effect of the workaround is that `LoadingScreen` is displayed in `WalletScreenTest`. The test isn't a `@HiltTest` but `LoadingScreen` has a default `hiltViewModel()` parameter which causes the test to fail.

So now `WalletScreenTest` can supply the view model for `LoadingScreen` instead of using `hiltViewModel()`.

alex-bradbury-gds (approved):

PR #1008, opened 2026-08-19
fix: DCMAW-22137 Handle requests from Wallet SDK to enter full screen

jonnyandrew:
## Context

Fixes an issue where Wallet SDK couldn't enter full screen.

This was happening because we were using two separate instances of `WalletSdk` to 1) display the `WalletApp()` composable and 2) listen for full screen requests.

To fix this, we listen for full screen requests from the same instance of `WalletSdk` that displays `WalletApp()`.

Another solution would have been to scope `WalletSdk` as a `@Singleton`, but I decided against this because we've discussed potentially migrating it away from `@SingletonScope` in future. Making it a `@Singleton` now could make that change harder over time.

## Evidence

[Screen_recording_20260819_155744.webm](https://github.com/user-attachments/assets/b4e74c8b-07f1-473a-a47d-a91cab4bff9a)


core/src/main/java/uk/gov/onelogin/core/ui/wallet/WalletDisplayer.kt
```diff
@@ -5,10 +5,12 @@ import androidx.compose.runtime.Composable
 /**
  * An interface for initialising the wallet sdk
  */
-fun interface WalletAppDisplayer {
+fun interface WalletDisplayer {
 
     @Composable
-    fun WalletApp()
+    fun Wallet(
```
jonnyandrew, line 11: I've renamed from `WalletApp` to `Wallet` to reflect that it is no longer simply a wrapper for the `WalletApp()` composable.

core/src/main/java/uk/gov/onelogin/core/ui/wallet/WalletDisplayerImpl.kt
```diff
@@ -0,0 +1,28 @@
...
+class WalletDisplayerImpl @Inject constructor(
+    private val walletSdk: WalletSdk
+) : WalletDisplayer {
+
+    @Composable
+    override fun Wallet(
+        onFullScreenRequest: (Boolean) -> Unit
+    ) {
+        val fullScreen by walletSdk
+            .displayAsFullScreen
```
jonnyandrew, line 19: Only the instance of `WalletSdk` that we show the `WalletApp()` composable with will emit `displayAsFullScreen` values.

Moving this into `WalletDisplayer` ensures that both `WalletApp()` and `displayAsFullScreen` are called on the same instance of `WalletSdk`.

Another solution would be to scope `WalletSdk` as a `@Singleton`, but I decided against this because we've discussed potentially migrating it away from `@SingletonScope` in future. Making it a `@Singleton` now could make that change harder over time.

elijahhezekiah-maker (approved):

PR #1015, opened 2026-08-26
test: Update Android test device API level from 33 to 35

jonnyandrew:


.sh/tests/runInstrumentationChecks
```diff
@@ -18,7 +18,7 @@ set -o errexit
 # Provide a default value for WORKER_COUNT
 if [[ -z "${WORKER_COUNT}" ]]
 then
-  WORKER_COUNT=2
+  WORKER_COUNT=1
```
jonnyandrew: I noticed that tests that are sharded started to fail recently and I'm not sure why. Reducing the worker count seems to stop the tests failing.

This doesn't seem to have an impact on the time spent running instrumentation tests (this PR takes 15 mins, a recent run on `develop` takes 16 mins).

elijahhezekiah-maker (approved):

build.gradle.kts
```diff
@@ -117,7 +117,7 @@ val emulatorConfig: EmulatorConfig by extra(
         systemImageSources = setOf(
             SystemImageSource.AOSP_ATD
         ),
-        androidApiLevels = setOf(33),
+        androidApiLevels = setOf(35),
```
alex-bradbury-gds, line 120: Is there a reason for not going up to 36?

jonnyandrew: Only that I'd thought it wasn't installed by default on the GitHub Actions runners.

But it turns out it is:

image

I'll raise another PR to update to 36

PR #1054, opened 2026-09-16
chore: DCMAW-23479 Update MOD certificate pinning config for production environment

jonnyandrew:


app/src/production/res/xml/network_security_config.xml
jonnyandrew, line 1: Note the file path - this is the production environment (app/src/production)

elijahhezekiah-maker (approved):

